PR #7, opened 2022-12-16
unify Layer model and Db model

vincentsarago:
This PR does:
- move all code from `tipg.layer` to `tipg.dbmodel` 
Now that Functions and Table are unified, there was no need of having a BaseClass and also no need to have a `layer.Table` class. Everything could be defined with one `Collection` class 

- rename `table_catalog` to `collection_catalog`
- `table` -> `collection` in multiple places


tipg/dbmodel.py
```diff
@@ -71,14 +140,17 @@ class Parameter(Column):
     default: Optional[str] = None
 
 
-class Table(BaseModel):
-    """Model for DB Table."""
+class Collection(BaseModel):
```
vincentsarago, line 143: Next add `from_file` class method to create a Collection from a SQL file

https://github.com/developmentseed/timvt/blob/master/timvt/layer.py#L252-L258

tipg/factory.py
```diff
@@ -382,10 +381,7 @@ def collections(
             output_type: Optional[MediaType] = Depends(OutputType),
         ):
             """List of collections."""
-            function_catalog = getattr(request.app.state, "tipg_function_catalog", {})
-            table_catalog = getattr(request.app.state, "table_catalog", {})
-            # convert all table to Table object
-            tables = [TableLayer(**t) for t in table_catalog.values()]
+            collection_catalog = getattr(request.app.state, "collection_catalog", {})
```
vincentsarago, line 384: For now let assume we will have only one catalog

tipg/dbmodel.py
```diff
@@ -446,19 +1068,19 @@ async def get_table_index(
...
-                "description": table["description"],
-                "id_column": id_column,
-                "geometry_columns": geometry_columns,
-                "datetime_columns": datetime_columns,
-                "properties": properties,
-                "datetime_column": datetime_column,
-                "geometry_column": geometry_column,
-                "parameters": table["parameters"],
-            }
+            catalog[id] = Collection(
```
vincentsarago, line 1071: We now store `Collection` object directly into the catalog (instead of a dictionary)

bitner (approved):

PR #13, opened 2023-01-03
get min/max for date columns

bitner:
Get min/max for datetime columns.

It's not entirely ideal because we need to make a request to the database for each datetime column, but without making a utility function similar to st_estimatedextent, I'm not sure we have much of an option.

### DIFF
--- a/tipg/dbmodel.py
+++ b/tipg/dbmodel.py
@@ -1044,16 +1044,15 @@ async def get_collection_index(  # noqa: C901
         fallback_key_names = table_settings.fallback_key_names
 
         for table in rows:
-            print(table)
-            id = table["id"]
+            table_id = table["id"]
 
-            if table_settings.excludes and id in table_settings.excludes:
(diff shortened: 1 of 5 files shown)


tipg/dbmodel.py
```diff
@@ -1048,11 +1049,26 @@ async def get_collection_index(
...
-                for c in table.get("datetime_columns", [])
-                if c["name"] in property_names
-            ]
+            datetime_columns = []
+            datetime_column = None
+            for c in table.get("datetime_columns", []):
+                print(c, property_names)
+                if c["name"] in property_names:
+                    # get min/max from database
+                    mindt, maxdt = await conn.fetchrow_b(
```
vincentsarago, line 1058: What do you think about the Settings to enable/disable this?

bitner: I think that would be reasonable. I would leave it on as the default.

PR #15, opened 2023-02-08
include layer/table name in MVT

NiekFi:
## What I am changing

- We wanted to include the queried layer name inside the MVT object for referencing in the front end.

## How I did it

- By passing the table name to the ST_AsMVT() function.


vincentsarago:
@NiekFi great addition, do you think you could add a test for this 🙏

NiekFi:
we tried to run the test for an hour but it seemed to fail in the same way as the automatic tests :sweat_smile: We'll try to include it

RemcoMeeuwissen:
We are seeing that some tests will fail (once/if we can run them properly haha) so we will try to fix that. You don't happen to know why the tests are failing with: sqlalchemy.exc.ObjectNotExecutableError: Not an executable object: 'CREATE EXTENSION  "postgis"'?

vincentsarago:
@NiekFi @RemcoMeeuwissen do the tests work locally? I think the CI is broken, I'll have a look

NiekFi:
no they do not, similar errors as the CI: sqlalchemy.exc.ObjectNotExecutableError: Not an executable object: 'CREATE EXTENSION "postgis"'?

vincentsarago:
ok I think it's because of the new sqlachemy (2.0) which just got released. let me try to fix this

vincentsarago:
@NiekFi can you pull the latest from `main` which includes the fix https://github.com/developmentseed/tipg/pull/16

bitner:
The problem with this is that many naive vector tile clients (including what we use in the templates for the /collection endpoint) that are not using style documents for vector tiles only look for the "default" layer. I would be willing to add a setting that allows you to chose in an environment variable whether to use "default" or <tablename> when creating vector tiles, but this change would at the very least break our templates as-is.

bitner (changes requested):
The problem with this is that many naive vector tile clients (including what we use in the templates for the /collection endpoint) that are not using style documents for vector tiles only look for the "default" layer. I would be willing to add a setting that allows you to chose in an environment variable whether to use "default" or when creating vector tiles, but this change would at the very least break our templates as-is.

RemcoMeeuwissen:
I gave the styling in the template a shot. It seems that geometryDimension paramter in collection.html was always undefined so none of the styling got applied before. I fixed this by reducing it to a style which gets applied to every type of geom (from my testing with the default data it looks fine).

The default/custom layer naming is fixed by setting the styling on both of these at the same time, with the custom layer name being taken from response.title. Did some manual tests and they got applied correctly with both use cases.

For the other part of the issue you raised, how about a .env setting for enabling the custom layer names?

vincentsarago:
@NiekFi @RemcoMeeuwissen we really appreciate the contribution, We are doing quite a refactor in #19 so we might wait for it to be done before continuing the work here. I hope you'll understand 🙏

RemcoMeeuwissen:
No worries, but if the .env idea sounds good we can implement that while you guys are working on #19 since it's a relatively small fix :smiley:

bitner:
+1 on making it comfortable by .env. Any new env bars should be setup in settings.py

bitner:
+1 on making it configurable by .env. Any new env bars should be setup in settings.py

bitner:
@NiekFi @RemcoMeeuwissen This is looking great! Thanks, especially, for reworking to deal with the larger changes in #19 and making sure to include tests! It looks like there are still some tests not passing, but I think once we get those all passing, we should be able to merge this pretty quickly.

bitner (dismissed):

RemcoMeeuwissen:
Right okay, so the tests should now also properly tests for both a default named layer and a custom named layer inside a MVT. I think that should be everything needed for this feature if the tests etc. pass :smiley:

RemcoMeeuwissen:
Guess using the merge tool on Github made Black unhappy, the suffering never ends haha. We'll run it locally tomorrow and do another push then.

bitner (approved):

bitner:
Thanks @RemcoMeeuwissen !

PR #19, opened 2023-02-08
Usepgtemp

bitner:
Use the pg_temp schema to create introspection functions to get metadata about tables, views, and functions.

Allow user created functions, views, and tables to be created in the pg_temp schema from a directory of sql files (set by the environment variable TIPG_FUNCTION_DIRECTORY) to be exposed by tipg.

Cleanup Column and Collections models to use property functions to calculate items that can be calculated (ie geometry columns, extent).


dockerfiles/Dockerfile
```diff
@@ -6,11 +6,12 @@ WORKDIR /tmp
 
 COPY README.md README.md
 COPY LICENSE LICENSE
-COPY tipg/ tipg/
 COPY pyproject.toml pyproject.toml
+RUN mkdir -p tipg && echo '__version__="0.0.0"' >/tmp/tipg/__init__.py
```
vincentsarago, line 10: 🤔 what's that for?

pyproject.toml
```diff
@@ -87,6 +87,10 @@ exclude = [
   ".gitignore",
   ".pre-commit-config.yaml",
 ]
+include = [
+  "/tipg/dbcatalog.sql",
+  "/tipg/templates/*.html",
+]
```
vincentsarago, line 93: that shouldn't be necessary

dockerfiles/Dockerfile
bitner: To install dependencies early in the docker chain so cache doesn't change

tipg/db.py
```diff
@@ -29,7 +31,8 @@ async def connect_to_db(
     """Connect."""
     if not settings:
         settings = PostgresSettings()
-
+    apisettings = APISettings()
```
vincentsarago, line 34: I think it would be better to pass the `functions_directory` as an optional input to the connect_to_db function or as PostgreSettings attribute instead of using the  `APISettings`

dockerfiles/Dockerfile
vincentsarago: ok, seems like pretty `hacky` I don't think we should do that.

If creating docker image speed is really an issue we could do something like

```
RUN pip install pip-tools
...
COPY pyproject.toml pyproject.toml
RUN pip-compile -o requirements.txt pyproject.toml
RUN pip install -r requirement.txt
```

But again, we shouldn't care about the speed of building the docker image.

tipg/dbcatalog.sql
```diff
@@ -0,0 +1,237 @@
+CREATE OR REPLACE FUNCTION pg_temp.typ(t text) RETURNS text AS $$
```
vincentsarago, line 1: let's move this into a `/sql` directory

tipg/dbcatalog.sql
```diff
@@ -0,0 +1,237 @@
...
+            AND (functions IS NULL OR proname::text = ANY (functions))
+    )
+    SELECT meta FROM a
+    WHERE
+        CASE
+            WHEN spatial THEN meta @? '$.properties[*] ? (exists (@.geometry_type))'
+            ELSE TRUE
+        END
+    ;
+$$ LANGUAGE SQL;
```
vincentsarago, line 237: ❤️

tipg/dbmodel.py
```diff
@@ -150,45 +171,43 @@ class Collection(BaseModel):
...
-            values["extent"] = Extent(**extent)
+    @property
+    def crs(self):
+        if self.geometry_column:
+            return f"http://www.opengis.net/def/crs/EPSG/0/{self.geometry_column.srid}"
 
-        return values
+    @property
+    def datetime_columns(self):
+        return [c for c in self.properties if c.is_datetime]
```
vincentsarago, line 210: ✨

tipg/dbmodel.py
```diff
@@ -309,23 +319,42 @@ def _select(
...
+            or llbounds.bottom < -85
+        ):
+            geom = logic.Func(
+                "st_intersection",
+                logic.Func("st_makeenvelope", -180, 85, 180, -85, 4326),
+                logic.Func(
+                    "st_transform",
+                    logic.V(geometry_column.name),
+                    pg_funcs.cast(4326, "int"),
+                ),
```
vincentsarago, line 342: Can you explain what's going on here?

tipg/dbmodel.py
```diff
@@ -489,9 +519,9 @@ def _where(
                             logic.Func(
                                 "ST_MakeEnvelope",
                                 tilebox.left,
-                                tilebox.bottom,
+                                max(-85, tilebox.bottom),
                                 tilebox.right,
-                                tilebox.top,
+                                min(85, tilebox.top),
```
vincentsarago, line 524: we could `truncate` the lat/lon directly with morecantile `truncate_lnglat` method (so it will work for all TMS)

tipg/dbmodel.py
bitner: It clips geometries at +-85 so that things don't error near the poles when converting to 3857.

tipg/dbmodel.py
bitner: :eyes: I did not know such a method existed!

vincentsarago: 👍 so I guess we can also use it at L335

tipg/db.py
```diff
@@ -11,14 +12,37 @@
 from fastapi import FastAPI
 
 
-async def con_init(conn):
+async def con_init(
+    conn,
+    settings: Optional[PostgresSettings] = None,
```
vincentsarago, line 17: This will make customization a bit hard, because we won't be able to pass any settings to it

tipg/db.py
```diff
@@ -11,14 +12,37 @@
...
+                'search_path',
+                'pg_temp,' || current_setting('search_path', false),
+                false
+                );
+            """
+    )
+    if (
+        settings.tipg_functions_directory
+        and Path(settings.tipg_functions_directory).exists()
+    ):
```
vincentsarago, line 41: we already make sure in the settings model that the directory exists

tipg/db.py
```diff
@@ -11,14 +12,37 @@
...
+        for sqlfile in Path(settings.tipg_functions_directory).glob("**/*.sql"):
+            await conn.execute(sqlfile.read_text())
+    dbcatalogsql = Path("tipg/sql/dbcatalog.sql")
```
vincentsarago, line 44: I think it should be `Path(__package__ / "sql" / "dbcatalog.sql")`

tipg/db.py
bitner: In order to use the temp schema, we need to make sure that the functions_directory scripts run on each and every connection. Doing it when creating the pool does not work.

vincentsarago: Sure, I'll think about something 

```python
from pydantic import BaseModel

class Conn(BaseModel):

    settings: PostgresSettings

    async def __call__(self, conn)
        await conn.set_type_codec(
            "json", encoder=orjson.dumps, decoder=orjson.loads, schema="pg_catalog"
        )
        await conn.set_type_codec(
            "jsonb", encoder=orjson.dumps, decoder=orjson.loads, schema="pg_catalog"
        )
        await conn.execute(
            """
                SELECT set_config(
                    'search_path',
                    'pg_temp,' || current_setting('search_path', false),
                    false
                    );
                """
        )
        if (
            self.settings.tipg_functions_directory
            and Path(self.settings.tipg_functions_directory).exists()
        ):
            for sqlfile in Path(self.settings.tipg_functions_directory).glob("**/*.sql"):
                await conn.execute(sqlfile.read_text())
        
        dbcatalogsql = Path(__package__ / "sql" / "dbcatalog.sql")
        
        await conn.execute(dbcatalogsql.read_text())
```

vincentsarago: but let me think about this more

tipg/dbmodel.py
```diff
@@ -150,45 +160,47 @@ class Collection(BaseModel):
...
+            spatial = Spatial(bbox=[self.geometry_column.bounds], crs=self.crs)
+        if (
+            self.datetime_column
+            and self.datetime_column.mindt
+            and self.datetime_column.maxdt
+        ):
+            temporal = Temporal(
+                interval=[[self.datetime_column.mindt, self.datetime_column.maxdt]]
+            )
+        return Extent(spatial=spatial, temporal=temporal)
```
vincentsarago, line 192: We should get the extent from all the columns (`self.geometry_columns` and `self.datetime_columns`) not only form `self.datetime_column` and `self.geometry_column`

vincentsarago: in fact for spatial, the bbox should be in form of `[[total bounds], [bounds of geom col1], [bounds of geom col2], ....]

> One or more bounding boxes that describe the spatial extent of the dataset.      
         Extensions may support additional areas.
          The first bounding box describes the overall spatial
          extent of the data. All subsequent bounding boxes describe
          more precise bounding boxes, e.g., to identify clusters of data.
          Clients only interested in the overall spatial extent will
          only need to access the first bounding box in the array.

tipg/factory.py
```diff
@@ -444,8 +447,9 @@ def collections(
                 collections_list = [
                     collection
                     for collection in collections_list
-                    if collection.bounds is not None
-                    and s_intersects(bbox_filter, collection.bounds)
+                    if collection.extent is not None
+                    and collection.extent.spatial is not None
+                    and s_intersects(bbox_filter, collection.extent.spatial.bbox)
```
vincentsarago, line 452: the collection's spatial extent is the sum of all the column extent so we can use `collection.bounds` here

tipg/dbmodel.py
vincentsarago: Same for temporal 
> The first time interval describes the overall
temporal extent of the data. All subsequent time intervals describe
more precise time intervals, e.g., to identify clusters of data.
Clients only interested in the overall temporal extent will only need
to access the first time interval in the array (a pair of lower and upper
bound instants).

tipg/factory.py
```diff
@@ -594,11 +598,11 @@ def collection(
             output_type: Optional[MediaType] = Depends(OutputType),
         ):
             """Metadata for a feature collection."""
+
             data = model.Collection(
                 **{
-                    "id": collection.id,
+                    **collection.dict(),
```
vincentsarago, line 604: This is a bit more `costly` because you ask pydantic to create a dict 🤷

tipg/dbmodel.py
bitner: I'm not sure what I was doing was either right or the best thing to do.

We can just skip doing anything for now, but I do think that we are going to want to do something for this since for a number of examples we end up getting all kinds of errors that bubble up from postgis when we do nothing. The approach I took here was really playing whack-a-mole with those problems rather than anything well thought out.

tipg/dbmodel.py
bitner: I think getting the overall bounds should be priority for this PR. We can ticket getting the separate bounds for columns to another issue.

vincentsarago: see #21

vincentsarago (approved):

vincentsarago:
I think we can merge this one, I'll add tests for the functions in another PR @bitner

PR #21, opened 2023-02-10
update SQL Directory env and collection Extent

vincentsarago:
This PR does:
- update mypy (couldn't get the all version run on my laptop 🙈)
- replace `TIPG_FUNCTIONS_DIRECTORY` by `TIPG_CUSTOM_SQL_DIRECTORY` 
- remove `TIPG_CUSTOM_SQL_DIRECTORY` from PostgresSettings
- refactor spatial and temporal extent (and fix tests)

### DIFF
--- a/tipg/db.py
+++ b/tipg/db.py
@@ -1,31 +1,36 @@
 """tipg.db: database events."""
 
-from pathlib import Path
+import os
+import pathlib
 from typing import Any, Optional
 
 import orjson
 from buildpg import asyncpg
 
 from tipg.dbmodel import get_collection_index
+from tipg.errors import FunctionDirectoryDoesNotExist
 from tipg.settings import PostgresSettings
 
 from fastapi import FastAPI
 
+custom_sql: list[pathlib.Path] = []
+user_sql_dir = os.environ.get("TIPG_CUSTOM_SQL_DIRECTORY", None)
+if user_sql_dir:
+    if not pathlib.Path(user_sql_dir).exists():
+        raise FunctionDirectoryDoesNotExist
 
-async def con_init(
-    conn,
-    settings: Optional[PostgresSettings] = None,
-):
-    """Use json for json returns."""
-    if not settings:
-        settings = PostgresSettings()
+    custom_sql = list(pathlib.Path(user_sql_dir).glob("*.sql"))
 
+
+async def con_init(conn):
+    """Use orjson encoder/decoder and register custom SQL functions."""
     await conn.set_type_codec(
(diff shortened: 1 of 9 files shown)


tipg/db.py
```diff
@@ -1,31 +1,36 @@
...
 
 from fastapi import FastAPI
 
+custom_sql: list[pathlib.Path] = []
+user_sql_dir = os.environ.get("TIPG_CUSTOM_SQL_DIRECTORY", None)
+if user_sql_dir:
+    if not pathlib.Path(user_sql_dir).exists():
+        raise FunctionDirectoryDoesNotExist
+
+    custom_sql = list(pathlib.Path(user_sql_dir).glob("*.sql"))
```
vincentsarago, line 22: This is just simpler, we didn't really need this to be in the `PostgresSettings` because it's more an TiPG settings.

tipg/dbmodel.py
```diff
@@ -135,9 +135,7 @@ def json_type(self) -> str:
     @property
     def is_geometry(self) -> bool:
         """Returns true if this property is a geometry column."""
-        if self.geometry_type:
-            return True
-        return False
+        return self.type in ("geometry", "geography")
```
vincentsarago, line 138: for consistency with `is_datetime`

tipg/factory.py
```diff
@@ -87,20 +87,17 @@ def write(self, line: str):
...
-            and (bbox[1] < ext[3])
-        ):
-            return True
-    return False
+    return (
+        (bbox[0] < spatial_extent[2])
+        and (bbox[2] > spatial_extent[0])
+        and (bbox[3] > spatial_extent[1])
+        and (bbox[1] < spatial_extent[3])
+    )
```
vincentsarago, line 97: we use the first interval which is the global interval

tipg/db.py
bitner: The approach is fine, but I think we should still do this somewhere in settings.py so that all the settings can be found in one place for anyone who is trying to look for all the settings they can use.

bitner (changes requested):
Only change here I would want is that I'd still like to see the TIPG_CUSTOM_SQL_DIRECTORY get defined somewhere (I don't care if it's in pgsettings or wherever) in settings.py so that someone looking to find the different settings they can use can always just go to that one place.

vincentsarago:
@bitner so latest commit refactored a bit the settings. we now have 3 Settings:
- PostgresSettings: Connection information
- DatabaseSettings: information about tables/schemas 
- CustomSQLSettings: path to the custom sql directory 

Let me know what you think

bitner (approved):

Issue #23, opened 2023-02-13
It's a bug not a feature! reproject if needed (not the opposite) 


vincentsarago:
https://github.com/developmentseed/tipg/blob/0c0f5129fc4405e7548db5e074926bdedd995983/tipg/dbmodel.py#L385-L386

cc @bitner

vincentsarago:
this was fixed in #24

PR #24, opened 2023-02-16
some geo related edits

vincentsarago:
This PR proposes some edits

### DIFF
--- a/tipg/dbmodel.py
+++ b/tipg/dbmodel.py
@@ -300,14 +300,14 @@ class Collection(BaseModel):
         geom = self._geom(geometry_column, bbox_only, simplify)
         if geom_as_wkt:
             if geom:
-                sel = sel.comma(logic.Func("st_asewkt", geom).as_("tipg_geom"))
+                sel = sel.comma(logic.Func("ST_AsEWKT", geom).as_("tipg_geom"))
             else:
                 sel = sel.comma(pg_funcs.cast(None, "text").as_("tipg_geom"))
 
         else:
             if geom:
                 sel = sel.comma(
-                    pg_funcs.cast(logic.Func("st_asgeojson", geom), "json").as_(
+                    pg_funcs.cast(logic.Func("ST_AsGeoJSON", geom), "json").as_(
                         "tipg_geom"
                     )
                 )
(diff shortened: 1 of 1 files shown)


tipg/dbmodel.py
```diff
@@ -409,14 +407,17 @@ def _geom(
         if geometry_column is None:
             return None
 
-        g = logic.V(geometry_column.name)
-        g = pg_funcs.cast(g, "geometry")
+        g = pg_funcs.cast(logic.V(geometry_column.name), "geometry")
 
-        if geometry_column.srid == 4326:
+        # Reproject to WGS64 if needed
+        if geometry_column.srid != 4326:
```
vincentsarago, line 413: fixes #23

tipg/dbmodel.py
```diff
@@ -507,24 +508,29 @@ def _where(
             wheres.append(to_filter(cql, [p.name for p in self.properties]))
 
         if tile and tms and geometry_column:
-            tilebox = tms.bounds(tile)
+            # Get Tile Bounds in Geographic CRS (usually epsg:4326)
+            left, bottom, right, top = tms.bounds(tile)
```
vincentsarago, line 512: One potential issue is that morecantile do not enforce the geographic CRS to be `epsg:4326`, we should maybe only accept TMS that uses WGS84. (note this is because morecantile is used with non-earth bodies)

tipg/dbmodel.py
```diff
@@ -507,24 +508,29 @@ def _where(
...
+
+            # Truncate bounds to the max TMS bbox
+            left, bottom = tms.truncate_lnglat(left, bottom)
+            right, top = tms.truncate_lnglat(right, top)
```
vincentsarago, line 516: to me it feels weird that we use bounds in geographic CRS instead of the bounds in TMS CRS. I guess the issue is that we cannot use `ST_MakeEnvelope` with non-epsg CRS 🤷

bitner: The intersects should be in the datasets srs since that is what has a spatial index on it.

vincentsarago: sure, but at this line we use `bounds` in WGS84 from morecantile to construct the envelope which is then Reprojected to the geometry CRS. My comment is mostly to say that we shouldn't rely on morecantile bounds in WGS84, but do the transformation and creation of the envelop in Postgres

bitner (approved):

PR #25, opened 2023-02-17
fix prop type

### LINKED ISSUE #14: `get_collection_index` fails with Postgres < 14
```
asyncpg.exceptions.UndefinedFunctionError: operator does not exist: text[] || name
HINT:  No operator matches the given name and argument types. You might need to add explicit type casts.
```

https://github.com/developmentseed/tipg/blob/51f4dd51d204945ca75ea35cf16cfa853397f9ce/tipg/dbmodel.py#L801-L1029

vincentsarago:
closes #14

### DIFF
--- a/tipg/db.py
+++ b/tipg/db.py
@@ -12,6 +12,12 @@ from tipg.settings import CustomSQLSettings, PostgresSettings
 
 from fastapi import FastAPI
 
+try:
+    from importlib.resources import files as resources_files  # type: ignore
+except ImportError:
+    # Try backported to PY<39 `importlib_resources`.
+    from importlib_resources import files as resources_files  # type: ignore
+
 sql_settings = CustomSQLSettings()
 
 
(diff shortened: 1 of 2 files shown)


vincentsarago:
Sadly it seems pretty hard to change our CI to test multiple version of Pg, you'll have to trust me for now 😅

tipg/db.py
```diff
@@ -48,7 +54,7 @@ async def con_init(conn):
             await conn.execute(sqlfile.read_text())
 
     # Register TiPG functions
-    dbcatalogsql = pathlib.Path(__package__) / "sql" / "dbcatalog.sql"
+    dbcatalogsql = resources_files(__package__) / "sql" / "dbcatalog.sql"
```
vincentsarago, line 57: When `pathlib.Path(__package__)` is returning a relative path, when using tipg from another application it will then error not finding the sql files.

bitner (approved):

Issue #26, opened 2023-02-17
remove vector tiles viz in `/collection` pages and add a `/collections/{collection}/map` viewer 


vincentsarago:
https://github.com/developmentseed/tipg/blob/4923d32abe434ddd549f41456c08b91d75762b65/tipg/templates/collection.html#L61-L91

When working with big collection, the viewer is not really useful, and removing this will also help splitting the features api and the tiles api.

cc @bitner

bitner:
:thinking: so... the biggest use case I had for having the vector tiles in the collection was just so I had an easy place that I could quickly just see that vector tiles were working. I definitely see your point though.

vincentsarago:
done in https://github.com/developmentseed/tipg/pull/28

PR #27, opened 2023-02-17
remove mvt viz in collection page

vincentsarago:
ref #26

### DIFF
--- a/tipg/templates/collection.html
+++ b/tipg/templates/collection.html
@@ -57,39 +57,7 @@
       [bbox[3], bbox[0]]
     ]);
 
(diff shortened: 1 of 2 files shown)


bitner:
Hrmmm. I totally get (and agree) with why you want to get rid of this, but it is really nice for debugging and convenience to have a simple viewer to check that vector tiles is working. Could we maybe add a /collections/{collection}/tiles endpoint that was just the template with the map frame?

vincentsarago:
>  Could we maybe add a /collections/{collection}/tiles endpoint that was just the template with the map frame?


I'll add this in another PR

PR #28, opened 2023-02-20
add viewer

vincentsarago:
ref #27 
Screenshot 2023-02-21 at 12 40 17 AM

### DIFF
--- a/tipg/factory.py
+++ b/tipg/factory.py
@@ -50,7 +50,7 @@ from fastapi.responses import ORJSONResponse
 
 from starlette.datastructures import QueryParams
 from starlette.requests import Request
-from starlette.responses import Response, StreamingResponse
+from starlette.responses import HTMLResponse, Response, StreamingResponse
 from starlette.templating import Jinja2Templates, _TemplateResponse
 
 tilesettings = TileSettings()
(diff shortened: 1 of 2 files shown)


tipg/factory.py
```diff
@@ -1188,7 +1188,7 @@ async def tile(
             return Response(bytes(tile), media_type=MediaType.mvt.value)
 
         @self.router.get(
-            "/collections/{collectionId}/{TileMatrixSetId}/tilejson.json",
+            "/collections/{collectionId}/{tileMatrixSetId}/tilejson.json",
```
vincentsarago, line 1191: should have added a test for this BUG

tipg/templates/map.html
```diff
@@ -0,0 +1,159 @@
+<!DOCTYPE html>
```
vincentsarago, line 1: This was the timvt viewer

tipg/factory.py
```diff
@@ -1251,17 +1251,72 @@ async def tilejson(
...
+        @self.router.get(
+            "/collections/{collectionId}/viewer", response_class=HTMLResponse
+        )
+        def viewer_endpoint(
+            request: Request,
+            collection=Depends(self.collection_dependency),
+            tileMatrixSetId: Literal["WebMercatorQuad"] = Query(  # noqa
+                "WebMercatorQuad",
+                description="Identifier selecting one of the TileMatrixSetId supported (default: 'WebMercatorQuad')",
+            ),
```
vincentsarago, line 1284: I choose to use the timvt viewer (using Maplibre) which only works with WebMercatorQuad TMS. The main issue is if a user will implement endpoints without WebMercatorQuad in the `supported_tms` 🤷

if we wanted to use leaflet + support multiple TMS we could use https://gist.github.com/vincentsarago/2fe69adda0b5089e9e6ca6bfe99bc7c9

bitner: I think maplibre is just fine. This viewer is mostly for quick testing/verifying.

bitner (approved):

Issue #30, opened 2023-02-24
Wrong CRS definition (epsg:4326 instead of CRS84)


vincentsarago:
in https://docs.opengeospatial.org/is/17-069r4/17-069r4.html#_coordinate_reference_systems 

> http://www.opengis.net/def/crs/OGC/1.3/CRS84 (WGS 84 longitude/latitude) for geometries without height information and http://www.opengis.net/def/crs/OGC/0/CRS84h (WGS 84 longitude/latitude plus ellipsoidal height) for geometries with height information.

ref: https://github.com/qgis/QGIS/pull/50468#issuecomment-1443832036

https://github.com/developmentseed/tipg/blob/9fbdb31fae359cd780957e6fd1eb3089fb2f20bb/tipg/model.py#L89

should be `http://www.opengis.net/def/crs/OGC/1.3/CRS84`

vincentsarago:
The other issue is then our spatial extent definition 

https://github.com/developmentseed/tipg/blob/9fbdb31fae359cd780957e6fd1eb3089fb2f20bb/tipg/dbmodel.py#L184-L187

IMO we should always have the `extent` in WGS84 (**lon,lat**) and set the extent CRS to `http://www.opengis.net/def/crs/OGC/1.3/CRS84` 

@bitner any comment on this?

vincentsarago:
closed in #31

PR #31, opened 2023-02-24
switch from epsg4326 to CRS84

### LINKED ISSUE #29: Filter bbox coordinate order needs to follow CRS
Could it be that the current `bbox` (bounding box) filter implementation in TiFeatures expects the parameters to be in order `xmin,ymin,xmax,ymax`?

We are experiencing an issue that a client (QGIS) is formatting the request such that **the order of the coordinate values follows the CRS**, which in the case of 4326 is `minlat,minlon,maxlat,maxlon`, i.e. `ymin,xmin,ymax,xmax`. 😩 

A request with the wrong order causes a 500 error with TiFeatures.

It appears that the `bbox` order needs to follow the CRS per WFS/OGC Features standard. I have not dug into the specs but found this in the [GeoServer docs](https://docs.geoserver.org/stable/en/user/services/wfs/reference.html):

> The BBOX parameter allows you to search for features that are contained (or partially contained) inside a box of user-defined coordinates. The format of the BBOX parameter is bbox=a1,b1,a2,b2,[crs] where a1, b1, a2, and b2 represent the coordinate values. The optional crs parameter is used to name the CRS for the bbox coordinates (if they are different to the featureTypes native CRS.) The order of coordinates passed to the BBOX parameter depends on the coordinate system used (this is why the coordinate syntax isn’t represented with x or y.)

Also, Even says so: https://github.com/qgis/QGIS/issues/33673#issuecomment-574791774 😉 

There seem to be many places in TiFeatures where `xmin,ymin,xmax,ymax` is hard-coded:

https://github.com/developmentseed/tifeatures/blob/324dd83a03b9620ef0b48564cad93930653bc7b2/tifeatures/filter/evaluate.py#L107

https://github.com/developmentseed/tifeatures/blob/5e0d9f319646fe09119aa8a2409fb71692337c70/tifeatures/dependencies.py#L165-L168

vincentsarago:
closes #29 , #30

### DIFF
--- a/tipg/dbmodel.py
+++ b/tipg/dbmodel.py
@@ -170,7 +170,7 @@ class Collection(BaseModel):
     @property
     def extent(self) -> Optional[Extent]:
         """Return extent."""
-        extent = {}
+        extent: Dict[str, Any] = {}
         if cols := self.geometry_columns:
             if len(cols) == 1:
                 bbox = [cols[0].bounds]
@@ -181,10 +181,7 @@ class Collection(BaseModel):
                     *[col.bounds for col in cols],
                 ]
 
-            extent["spatial"] = {
-                "bbox": bbox,
-                "crs": f"http://www.opengis.net/def/crs/EPSG/0/{cols[0].srid}",
-            }
+            extent["spatial"] = {"bbox": bbox}
 
         if cols := self.datetime_columns:
             cols = [col for col in cols if col.mindt or col.maxdt]
--- a/tipg/model.py
+++ b/tipg/model.py
@@ -42,7 +42,7 @@ class Spatial(BaseModel):
     # extent of the data. All subsequent bounding boxes describe
     # more precise bounding boxes, e.g., to identify clusters of data.
     bbox: List[List[float]]
-    crs: str
+    crs: str = "http://www.opengis.net/def/crs/OGC/1.3/CRS84"
 
 
 class Temporal(BaseModel):
@@ -86,7 +86,7 @@ class Collection(BaseModel):
     links: List[Link]
     extent: Optional[Extent]
     itemType: str = "feature"
-    crs: List[str] = ["http://www.opengis.net/def/crs/EPSG/0/4326"]
+    crs: List[str] = ["http://www.opengis.net/def/crs/OGC/1.3/CRS84"]
 
     class Config:
         """Collection model configuration."""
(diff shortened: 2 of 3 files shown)


tipg/dbmodel.py
```diff
@@ -181,10 +181,7 @@ def extent(self) -> Optional[Extent]:
                     *[col.bounds for col in cols],
                 ]
 
-            extent["spatial"] = {
-                "bbox": bbox,
-                "crs": f"http://www.opengis.net/def/crs/EPSG/0/{cols[0].srid}",
-            }
+            extent["spatial"] = {"bbox": bbox}
```
vincentsarago, line 184: from https://github.com/developmentseed/tipg/blob/9fbdb31fae359cd780957e6fd1eb3089fb2f20bb/tipg/sql/dbcatalog.sql#L82-L86 it seems that we are always returning bounds in WGS84 (as lon/lat)

cc @bitner

tipg/model.py
```diff
@@ -86,7 +86,7 @@ class Collection(BaseModel):
     links: List[Link]
     extent: Optional[Extent]
     itemType: str = "feature"
-    crs: List[str] = ["http://www.opengis.net/def/crs/EPSG/0/4326"]
+    crs: List[str] = ["http://www.opengis.net/def/crs/OGC/1.3/CRS84"]
```
vincentsarago, line 89: the default should be `CRS84`. Now the issue is that PostGIS will have the srid of the collection set to `4326` but all the input/output are in WGS84 (**lon/lat**) 🤷‍♂️

vincentsarago, line 89: The CRS property of the collection is `the list of coordinate reference systems supported by the service` so in TiPG it makes senses to have this set to `http://www.opengis.net/def/crs/OGC/1.3/CRS84` 

>   crs:
    description: the list of coordinate reference systems supported by the service
    type: array
    items:
      type: string
    default:
      - http://www.opengis.net/def/crs/OGC/1.3/CRS84
    example:
      - http://www.opengis.net/def/crs/OGC/1.3/CRS84
      - http://www.opengis.net/def/crs/EPSG/0/4326

ref: http://schemas.opengis.net/ogcapi/features/part1/1.0/openapi/schemas/collection.yaml

tipg/model.py
```diff
@@ -42,7 +42,7 @@ class Spatial(BaseModel):
     # extent of the data. All subsequent bounding boxes describe
     # more precise bounding boxes, e.g., to identify clusters of data.
     bbox: List[List[float]]
-    crs: str
+    crs: str = "http://www.opengis.net/def/crs/OGC/1.3/CRS84"
```
vincentsarago, line 45: we could leave this blank 🤷

vincentsarago:
@j08lue I'm not sure if you will be able to tests this branch but this should resolve the issue for your user. I'll update the code in tifeatures too once this gets merged!

bitner (approved):

PR #32, opened 2023-02-28
split OGC Features and Tiles endpoints factories

### LINKED ISSUE #6: split Features and Tiles endpoints in the Factory
let's have a `Endpoints().register_ogc_features()` and `Endpoints().register_ogc_tiles()` 

then we can do 

```python
    def __post_init__(self):
        """Post Init: register route and configure specific options."""
        self.register_landing()
        self.register_conformance()

        if self.with_ogc_features:
            self.register_ogc_features()
       
        if self.with_ogc_tiles:
            self.register_ogc_tiles()
```

vincentsarago:
closes #6 

This PR split the endpoint factory in two: `OGCFeaturesFactory` and `OGCTilesFactory` which should enable user to build either the tiles service or features service individually.

The main issue is for creating the OGC features `landing` and `conformances` endpoints. I've added two ways:
- within the `OGCFeaturesFactory`, by using the `full=True` option, this will add the `/` (landing) and `/conformance` endpoints but only referencing the features endpoint (no links to the Tiles endpoints).
    ```python
    from tipg.factory import OGCFeaturesFactory
    from fastapi import FastAPI

    app = FastAPI(
        openapi_url="/api",
        docs_url="/api.html",
    )
    features = OGCFeaturesFactory(full=True)
    app.include_router(features, tags=["OGC Features API"])
    ```

- in main.py, to have both tiles and features endpoints conformances and links. It is likely that a user who will want to add both `/` (landing), `/conformance` with tiles and features endpoints in it's own application will have to re-write both endpoints in its code 🤷‍♂️ 

IMO it's really annoying that both `/` and `/conformance` are part of the features specification 😬  

Screenshot 2023-02-28 at 9 47 15 AM

### DIFF
--- a/tipg/factory.py
+++ b/tipg/factory.py
@@ -1,5 +1,6 @@
 """tipg.factory: router factories."""
 
+import abc
 import csv
 from dataclasses import dataclass, field
 from typing import (
@@ -126,8 +127,52 @@ def t_intersects(interval: List[str], temporal_extent: List[str]) -> bool:
     return False
 
 
-@dataclass
-class Endpoints:
+def create_html_response(
+    request: Request,
+    data: str,
+    templates: Jinja2Templates,
+    template_name: str,
+    router_prefix: Optional[str] = None,
+) -> _TemplateResponse:
+    """Create Template response."""
+    urlpath = request.url.path
+    crumbs = []
+    baseurl = str(request.base_url).rstrip("/")
+
+    crumbpath = str(baseurl)
+    for crumb in urlpath.split("/"):
+        crumbpath = crumbpath.rstrip("/")
+        part = crumb
+        if part is None or part == "":
+            part = "Home"
+        crumbpath += f"/{crumb}"
+        crumbs.append({"url": crumbpath.rstrip("/"), "part": part.capitalize()})
+
+    if router_prefix:
+        baseurl += router_prefix
+
+    return templates.TemplateResponse(
+        f"{template_name}.html",
+        {
+            "request": request,
+            "response": orjson.loads(data),
+            "template": {
+                "api_root": baseurl,
+                "params": request.query_params,
+                "title": "",
+            },
+            "crumbs": crumbs,
+            "url": str(request.url),
+            "baseurl": baseurl,
+            "urlpath": str(request.url.path),
+            "urlparams": str(request.url.query),
+        },
+    )
+
+
+# ref: https://github.com/python/mypy/issues/5374
+@dataclass  # type: ignore
+class EndpointsFactory(metaclass=abc.ABCMeta):
     """Endpoints Factory."""
 
     # FastAPI router
@@ -140,19 +185,11 @@ class Endpoints:
     # e.g if you mount the route with `/foo` prefix, set router_prefix to foo
     router_prefix: str = ""
 
-    title: str = "TiPG API"
-
     templates: Jinja2Templates = DEFAULT_TEMPLATES
-
-    # OGC Tiles dependency
-    supported_tms: TileMatrixSets = default_tms
 
     def __post_init__(self):
         """Post Init: register route and configure specific options."""
-        self.register_landing()
-        self.register_conformance()
-        self.register_collections()
-        self.register_tiles()
+        self.register_routes()
 
     def url_for(self, request: Request, name: str, **path_params: Any) -> str:
         """Return full url (with prefix) for a specific handler."""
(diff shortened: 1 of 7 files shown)


tipg/factory.py
```diff
@@ -126,8 +127,52 @@ def t_intersects(interval: List[str], temporal_extent: List[str]) -> bool:
...
+                "params": request.query_params,
+                "title": "",
+            },
+            "crumbs": crumbs,
+            "url": str(request.url),
+            "baseurl": baseurl,
+            "urlpath": str(request.url.path),
+            "urlparams": str(request.url.query),
+        },
+    )
```
vincentsarago, line 170: we create a simple function that can be used outside the factories

tipg/factory.py
```diff
@@ -170,236 +207,46 @@ def _create_html_response(
...
-                return self._create_html_response(
-                    request,
-                    data.json(exclude_none=True),
-                    template_name="conformance",
-                )
+    @property
+    @abc.abstractmethod
+    def conforms_to(self) -> List[str]:
+        """Endpoints conformances."""
+        ...
```
vincentsarago, line 227: each Factory should explicitly tell what they are conform to

tipg/factory.py
```diff
@@ -467,11 +314,6 @@ def collections(
             items_returned = len(collections_list)
 
             links: list = [
-                model.Link(
-                    href=self.url_for(request, "landing"),
-                    rel="parent",
-                    type=MediaType.json,
-                ),
```
vincentsarago, line 474: `parent` is not a requirement. it's easier to get rid of it

vincentsarago:
@bitner any of this make sense? 😄

tipg/main.py
```diff
@@ -44,8 +48,190 @@
 )  # type: ignore
 
 # Register endpoints.
-endpoints = Endpoints(title=settings.name, templates=templates)
-app.include_router(endpoints.router, tags=["OGC API"])
+ogc_features = OGCFeaturesFactory(templates=templates)
+app.include_router(ogc_features.router, tags=["OGC Features API"])
+
+ogc_tiles = OGCTilesFactory(templates=templates)
+app.include_router(ogc_tiles.router, tags=["OGC Tiles API"])
```
vincentsarago, line 55: We create the endpoints using the 2 factories

tipg/main.py
```diff
@@ -44,8 +48,190 @@
...
+
+    if output_type == MediaType.html:
+        return create_html_response(
+            request,
+            data.json(exclude_none=True),
+            templates=templates,
+            template_name="conformance",
+        )
+
+    return data
```
vincentsarago, line 99: We hard code both conformance and landing but there is a method in the OGCFeaturesFactory if a user wants to create a standalone OGC Features API (with all the required endpoints)

vincentsarago:
latest version of FastAPI/starlette breaks the `url_for` methods https://github.com/encode/starlette/pull/2050 🤷

vincentsarago:
😬 it's the output of `url_for` which was a string before and now is of type URLpath

tipg/main.py
bitner: should we just have a single conformance/landing factory instead?

tipg/factory.py
```diff
@@ -1364,3 +1390,220 @@ async def tilematrixset(
...
+class Endpoints(EndpointsFactory):
+    """OGC Features and Tiles Endpoints Factory."""
+
+    title: str = "TiPg: OGC Features and Tiles API"
+
+    # OGC Tiles dependency
+    supported_tms: TileMatrixSets = default_tms
+
+    ogc_features: OGCFeaturesFactory = field(init=False)
+    ogc_tiles: OGCTilesFactory = field(init=False)
```
vincentsarago, line 1405: This factory combines both `OGCFeaturesFactory` and `OGCTilesFactory` factories.

tipg/factory.py
```diff
@@ -1364,3 +1390,220 @@ async def tilematrixset(
...
+
+    def __post_init__(self):
+        """Post Init: register route and configure specific options."""
+        self.ogc_features = OGCFeaturesFactory(
+            router=self.router,
```
vincentsarago, line 1410: by passing `self.router` to the Factory, it will register the endpoint to the `Endpoints.router`

tipg/factory.py
```diff
@@ -1364,3 +1390,220 @@ async def tilematrixset(
...
+            collection_dependency=self.collection_dependency,
+            router_prefix=self.router_prefix,
+            templates=self.templates,
+            full=False,
+            tags=["OGC Features API"],
```
vincentsarago, line 1415: we set Tags for the set of endpoints

tipg/factory.py
```diff
@@ -1364,3 +1391,359 @@ async def tilematrixset(
             OGC Specification: http://docs.opengeospatial.org/per/19-069.html#_tilematrixset
             """
             return self.supported_tms.get(tileMatrixSetId)
+
+        # NOTE: The OGC Tiles API specification includes `/conformances` and `/` endpoints
+        # Those two endpoints are optional in the factory and only added if `full=True`.
+        if self.full:
+
+            @self.router.get(
+                "/conformance",
```
vincentsarago, line 1400: `conformance` and `/` are OGC core specification so in theory we could create an OGC Tiles API with them + the tiles endpoints

tipg/factory.py
```diff
@@ -1392,358 +1424,49 @@ async def tilematrixset(
...
-            if output_type == MediaType.html:
-                return create_html_response(
-                    request,
-                    data.json(exclude_none=True),
-                    templates=self.templates,
-                    template_name="conformance",
-                )
-
-            return data
+        self.router.include_router(self.ogc_features.router, tags=["OGC Features API"])
```
vincentsarago, line 1462: we now use `include_router` method (without the prefix)

tipg/factory.py
```diff
@@ -1392,358 +1424,49 @@ async def tilematrixset(
...
-            if output_type == MediaType.html:
-                return create_html_response(
-                    request,
-                    data.json(exclude_none=True),
-                    templates=self.templates,
-                    template_name="landing",
-                )
-
-            return data
+        self.router.include_router(self.ogc_tiles.router, tags=["OGC Tiles API"])
```
vincentsarago, line 1472: -160 lines 💥

vincentsarago:
last commit refactored a bit how we add the `OGC common` endpoints. It's now defined at the base class level.

### To Do 
- [ ] add tests for each Factory

tipg/factory.py
```diff
@@ -1121,8 +1126,61 @@ async def item(
...
+                    "tilematrixset",
+                    tileMatrixSetId="{tileMatrixSetId}",
+                ),
+                type=MediaType.json,
+                rel="data",
+            ),
+        ]
+
+    def register_routes(self):  # noqa: C901
+        """Register OGC Features endpoints."""
```
bitner, line 1183: """Register OGC Tiles endpoints.""" ???

vincentsarago, line 1183:
```suggestion
        """Register OGC Tiles endpoints."""
```

bitner (approved):

bitner (approved):

PR #48, opened 2023-03-28
update search_path with user schemas

### LINKED ISSUE #44: Connect to alternative schemas
Hi, 

I'm running postgres, and have all my geospatial data in a schema "published" rather than using public. 

I'm struggling to connect to that schema - tipg is only connecting to public.

In the Dockerfile I've tried 

ENV POSTGRES_DBNAME=oraetl?options=-c%20search_path=published
and
ENV TIPG_DB_SCHEMAS=["published"]

Neither have worked. Any suggestions?

vincentsarago:
closes #44 

~~This PR tries to implement a solution based on https://github.com/developmentseed/tipg/issues/44#issuecomment-1486973429 but local tests shows that it doesn't work~~

```
# Launch database
docker-compose up database -d 

# Create a copy of `public.countries` in `myschema.countries`
psql postgresql://username:password@0.0.0.0:5439/postgis -c "CREATE SCHEMA IF NOT EXISTS myschema; SET schema 'myschema'; DROP TABLE IF EXISTS myschema.countries CASCADE; CREATE TABLE myschema.countries AS SELECT * FROM public.countries; SELECT count(*) FROM myschema.countries;"

# Start TiPG service
DATABASE_URL=postgresql://username:password@0.0.0.0:5439/postgis TIPG_DB_SCHEMAS='["myschema"]'  uvicorn tipg.main:app --port 8000 --reload
```

~~when `function_schemas` is set to `[]` using `TIPG_DB_FUNCTION_SCHEMAS='[]' then it works.~~

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -5,10 +5,15 @@ import os
 import pytest
 import pytest_pgsql
 
+from tipg.settings import CustomSQLSettings, DatabaseSettings, PostgresSettings
+
+from fastapi import FastAPI
+
 from starlette.testclient import TestClient
 
-DATA_DIR = os.path.join(os.path.dirname(__file__), "fixtures")
-
+FIXTURES_DIR = os.path.join(os.path.dirname(__file__), "fixtures")
+TEMPLATE_DIRECTORY = os.path.join(FIXTURES_DIR, "templates")
+SQL_FUNCTIONS_DIRECTORY = os.path.join(FIXTURES_DIR, "functions")
 
 test_db = pytest_pgsql.TransactedPostgreSQLTestDB.create_fixture(
     "test_db", scope="session", use_restore_state=False
@@ -27,13 +32,13 @@ def database_url(test_db):
     # make sure we have a `public` schema
     test_db.create_schema("public", exists_ok=True)
 
-    test_db.run_sql_file(os.path.join(DATA_DIR, "landsat_wrs.sql"))
+    test_db.run_sql_file(os.path.join(FIXTURES_DIR, "landsat_wrs.sql"))
     assert test_db.has_table("public.landsat_wrs")
 
-    test_db.run_sql_file(os.path.join(DATA_DIR, "my_data.sql"))
+    test_db.run_sql_file(os.path.join(FIXTURES_DIR, "my_data.sql"))
     assert test_db.has_table("public.my_data")
 
-    test_db.run_sql_file(os.path.join(DATA_DIR, "nongeo_data.sql"))
+    test_db.run_sql_file(os.path.join(FIXTURES_DIR, "nongeo_data.sql"))
     assert test_db.has_table("public.nongeo_data")
 
     test_db.connection.execute(
@@ -51,11 +56,11 @@ def database_url(test_db):
     assert count_landsat == count_landsat_centroid
 
     # Add table with Huge geometries
-    test_db.run_sql_file(os.path.join(DATA_DIR, "canada.sql"))
+    test_db.run_sql_file(os.path.join(FIXTURES_DIR, "canada.sql"))
     assert test_db.has_table("public.canada")
 
     # add table with geometries not in WGS84
-    test_db.run_sql_file(os.path.join(DATA_DIR, "minnesota.sql"))
+    test_db.run_sql_file(os.path.join(FIXTURES_DIR, "minnesota.sql"))
     assert test_db.has_table("public.minnesota")
 
     # add a `myschema` schema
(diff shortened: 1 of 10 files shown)


tipg/db.py
```diff
@@ -29,38 +29,48 @@
...
+
+    async def __call__(self, conn: asyncpg.Connection):
+        """Create connection."""
+        await conn.set_type_codec(
+            "json", encoder=orjson.dumps, decoder=orjson.loads, schema="pg_catalog"
+        )
+        await conn.set_type_codec(
+            "jsonb", encoder=orjson.dumps, decoder=orjson.loads, schema="pg_catalog"
+        )
+        schemas = ",".join(["pg_temp"] + self.schemas)
```
vincentsarago, line 49: The order of the path seems important.

cc @bitner

vincentsarago:
🤔 tests pass locally but not in CI

I'll wait for a pre-review from @bitner now 🙏 😄

tipg/db.py
bitner: We want to make sure that pg_temp is the first so that any new functions/views/tables created when we load sql from files are created into the pg_temp schema - if you don't specify a schema, new objects are written into the first schema in the search_path. We should also make sure to add the function_schemas from the environment as well.

bitner: Ahh never mind that last part of the comment, I see now below that self.schemas is already combined.

vincentsarago:
again after last commit, tests pass locally but I'm having hard time being able to reproduce the tests (it fails sometimes 🤷 ) 

I might update the `fixtures` to make sure we get 💯 reproductible tests

tests/conftest.py
```diff
@@ -71,17 +76,68 @@ def database_url(test_db):
...
+            exclude_functions=db_settings.exclude_functions,
+            spatial=db_settings.only_spatial_tables,
+        )
+
+    @app.on_event("shutdown")
+    async def shutdown_event() -> None:
+        """Close database connection."""
+        await close_db_connection(app)
+
+    return app
```
vincentsarago, line 127: our tests suite is a bit complex because we want to test our main application (`tipg.main.app`) but also test some customization. Sadly we have a lot of things that get initialized on `import` when importing `tipg.main.app` so this little help tries to create an application similar (but enable customization)

tipg/db.py
KoalaGeo: In other projects (pygeoapi) it's needed to be `search_path: [published, public]` if you used `search_path: [public, published]` it wouldn't pick up `published`

vincentsarago: Oo so I found why I can't make `reproductible` test, because I use `Set` https://github.com/developmentseed/tipg/pull/48/files#diff-772cfc11cca1ff02c30e13c6c3aa3ae382b238719d6eaaa11fa1df9c864e0a1fR163 then the order of schemas changes 😬

tipg/settings.py
```diff
@@ -140,13 +141,12 @@ class DatabaseSettings(pydantic.BaseSettings):
...
     schemas: List[str] = ["public"]
-    exclude_schemas: Optional[List[str]]
     tables: Optional[List[str]]
     exclude_tables: Optional[List[str]]
-    function_schemas: List[str] = ["public"]
-    exclude_function_schemas: Optional[List[str]]
+    exclude_table_schemas: Optional[List[str]]
     functions: Optional[List[str]]
     exclude_functions: Optional[List[str]]
+    exclude_function_schemas: Optional[List[str]]
```
vincentsarago, line 149: **breaking change** 

we now have only one `schemas` option for both `table` and `functions`. To exclude schemas for table or function we can use `exclude_table_schemas` or `exclude_function_schemas` 

cc @bitner

vincentsarago: we should also add a note somewhere about the importance of the schemas order. like `["myschema", "public"]` do not gives the same result as `["public", "myschema"]`

tipg/db.py
vincentsarago: fixed by unifying the schemas options into one (for both tables and functions)

tipg/sql/dbcatalog.sql
```diff
@@ -176,7 +184,7 @@ BEGIN
         jsonb_agg(json_strip_nulls(json_build_object(
             'name', proargname,
             'type', argtype,
-            'default', regexp_replace(def, '''([a-zA-Z0-9_\-\.]+)''::\w+', '\1', 'g')
```
bitner, line 179: This expression was not getting the defaults for functions populated correctly. Replaced this regex with the tipg_fun_defaults function.

tipg/sql/dbcatalog.sql
```diff
@@ -128,7 +128,15 @@ CREATE OR REPLACE FUNCTION pg_temp.tipg_tproperties(
     SELECT pg_temp.tipg_tproperties(pg_class) FROM pg_class WHERE oid=tabl::regclass;
 $$ LANGUAGE SQL;
 
-
+CREATE OR REPLACE FUNCTION pg_temp.tipg_fun_defaults(defaults pg_node_tree) RETURNS text[] AS $$
```
vincentsarago, line 131: `fun_defaults` ?

vincentsarago: you mean `func_defaults` maybe

tipg/sql/dbcatalog.sql
```diff
@@ -195,6 +203,29 @@ CREATE OR REPLACE FUNCTION pg_temp.tipg_fproperties(
     SELECT pg_temp.tipg_fproperties(pg_proc) FROM pg_proc WHERE oid=func::regproc;
 $$ LANGUAGE SQL;
 
+CREATE OR REPLACE FUNCTION pg_temp.tipg_get_schemas(include text[] DEFAULT NULL, exclude text[] DEFAULT NULL) RETURNS SETOF oid AS $$
```
bitner, line 206: Create a function that will take use the search_path if the passed in include schemas is null or empty otherwise it will use the include schemas and will exclude those from the exclude schemas as well as any schemas the user does not have permissions for.

tipg/sql/dbcatalog.sql
bitner: I think they are pretty fun!

tipg/settings.py
bitner: With the commit adding the tipg_get_schemas function, the order of schemas does not matter anymore. The call to pg_table_isvisible meant that for the "landsat" table that was in both public and myschema schemas, it would only ever return one of them into the catalog.

vincentsarago:
@KoalaGeo let's open another issue for the `decimal` error 🙏

vincentsarago:
🙏 thanks @bitner and @KoalaGeo

PR #64, opened 2023-04-14
switch to z/x/y tile parameter

### LINKED ISSUE #63: Switch from `{tileMatrix}/{tileRow}/{tileCol}` to `{z}/{x}/{y}`
> The tiles making up a tileset containing available data SHALL be available via an HTTP GET request to a URI that can be built from a template containing two or three variables. If the tileset is available for more than one tilematrix, the template SHALL contain three variables (such as {tileMatrix}, {tileRow} and {tileCol} as defined by the tileset conformance class, or {z}, {y} and {x}). If the tileset is available for a single tilematrix, the template SHALL contain two variables (such as {tileRow} and {tileCol} as defined by the tileset conformance class, or {y} and {x}). The URI is obtained by substituting the variables by their respective valid values.

ref: https://docs.ogc.org/is/20-057/20-057.html#toc28

Most of the map client (Mapbox/Maplibre) will understand `{z}/{x}/{y}` but not `{tileMatrix}/{tileRow}/{tileCol}` (https://github.com/developmentseed/tipg/blob/main/tipg/templates/map.html#L75)

vincentsarago:
closes #63

### DIFF
--- a/docs/src/advanced/factories.md
+++ b/docs/src/advanced/factories.md
@@ -116,7 +116,7 @@ app.include_router(endpoints.router, tags=["OGC Tiles API"])
 
 | Method | Path                                                                                     | Output                        | Description
 | ------ | ---------------------------------------------------------------------------------------- |------------------------------ |--------------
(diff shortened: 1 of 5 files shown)


bitner (approved):
I like this muuuuuch better.

PR #65, opened 2023-04-18
update conformance links for those that exist

ranchodeluxe:
## What I am changing

Fixing/removing conformance links that are 404'ing

## How I did it

I didn't immediately notice but the newest OGC docs (see links below) have removed the bad links, so I followed that example (outside of the ogcapi-features-3.0 examples which are broken and not removed).

https://docs.ogc.org/is/19-072/19-072.html#toc64
https://docs.opengeospatial.org/is/17-069r4/17-069r4.html#_response_3

I also left some questions on the relevant ogcapi repos:

https://github.com/opengeospatial/ogcapi-features/issues/821
https://github.com/opengeospatial/ogcapi-common/issues/327

## How you can test it

```bash
for link in links:
    curl -XGET link
```

### DIFF
--- a/tipg/factory.py
+++ b/tipg/factory.py
@@ -265,13 +265,11 @@ class EndpointsFactory(metaclass=abc.ABCMeta):
             """Get conformance."""
             data = model.Conformance(
                 conformsTo=[
-                    "http://www.opengis.net/spec/ogcapi-common-1/1.0/conf/core",
-                    "http://www.opengis.net/spec/ogcapi-common-1/1.0/conf/landingPage",
-                    "http://www.opengis.net/spec/ogcapi-common-2/1.0/conf/collections",
-                    "http://www.opengis.net/spec/ogcapi-common-2/1.0/conf/simple-query",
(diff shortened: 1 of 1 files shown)


ranchodeluxe:
> progress: [opengeospatial/ogcapi-features#822](https://github.com/opengeospatial/ogcapi-features/pull/822)
> 
> I'll wait to hear back from the `ogcapi-common` repo for fixes and see what they've done there. Then I can tweak things around and fix the tests here

Given the feedback on those ogcapi repo tickets we learned one big thing:
* that the OGC has embedded extra meaning into what a 404 means from the domain "http://www.opengis.net". They expect 404 users to somehow know, "oh, this must mean that the OGC spec this link is referring to isn't  an 'approved standard' yet"

I think that's pretty awful 😆 but I don't want to die on any OGC-redirect-link hills. This is their problem. So not going to make any changes here and let them figure it out

vincentsarago:
Thanks @ranchodeluxe 🙏 
Yeah OGC is sometime hard to follow ;-)

PR #66, opened 2023-05-03
Add TTL for refreshing catalog

bitner:
Start of work to add a TTL setting to force refetching the catalog for each collection when the entry for that collection is older than the TTL. When starting up, it always grabs all collections. When hitting the /collections endpoint, it will check each collection and only fetch ones older than the TTL. When hitting any /collections/{collection}/... endpoint, it will check that collection and only request that collection if older than the TTL.

TODO:

- [ ] Add parameter to force refetching.
- [ ] Tests

### DIFF
--- a/tipg/db.py
+++ b/tipg/db.py
@@ -1,14 +1,14 @@
 """tipg.db: database events."""
 
 import pathlib
-from typing import Any, List, Optional
+from typing import List, Optional
 
 import orjson
 from buildpg import asyncpg
 
 from tipg.dbmodel import get_collection_index
 from tipg.logger import logger
-from tipg.settings import PostgresSettings
+from tipg.settings import DatabaseSettings, PostgresSettings
 
 from fastapi import FastAPI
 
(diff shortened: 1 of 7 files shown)


tipg/db.py
```diff
@@ -70,15 +70,13 @@ async def __call__(self, conn: asyncpg.Connection):
 
 async def connect_to_db(
     app: FastAPI,
-    settings: Optional[PostgresSettings] = None,
+    settings: Optional[PostgresSettings] = PostgresSettings(),
```
vincentsarago, line 73: We should avoid!

tipg/dbmodel.py
```diff
@@ -164,6 +167,7 @@ class Collection(BaseModel):
     geometry_column: Optional[Column]
     datetime_column: Optional[Column]
     parameters: List[Parameter] = []
+    last_catalog_update: datetime = datetime.now()
```
vincentsarago, line 170: default should be set on the post validation step! the issue here is that you'll create a default when we import the dbmodel submodule

tipg/logger.py
```diff
@@ -2,4 +2,5 @@
 
 import logging
 
-logger = logging.getLogger("tipg")
+logger = logging.getLogger("uvicorn")
+logger.setLevel(logging.DEBUG)
```
vincentsarago, line 6: I know this is the easiest way to get the log in uvicorn but we shouldn't do this 😅 

see https://github.com/developmentseed/titiler/issues/276#issuecomment-806250688

tipg/db.py
bitner: What is wrong with this?

tipg/logger.py
bitner: It's staying in there on this branch as long as this PR is marked draft!

tipg/db.py
vincentsarago: we should not use mutable in function parameters

tipg/logger.py
vincentsarago: 🙈

bitner:
Closing in favor of #68

PR #68, opened 2023-05-04
approach using background tasks

bitner:
A different approach to updating the Catalog - this approach minimizes any wait in a request by refreshing the catalog as a background task so that it is updated for the next request.

This approach uses middleware to check the TTL setting as well as look for a refresh=true query parameter. 

If the refresh=true parameter is set, it will refresh the catalog and then proceed with the request.

If the TTL is expired, it will still use the cached catalog for this request, but it will submit a background task to update the catalog.

### DIFF
--- a/tipg/middleware.py
+++ b/tipg/middleware.py
@@ -1,8 +1,13 @@
 """tipg middlewares."""
 
 import re
+from datetime import datetime
 from typing import Optional, Set
 
+from tipg.db import register_collection_catalog
+from tipg.logger import logger
+
+from starlette.background import BackgroundTask
 from starlette.middleware.base import BaseHTTPMiddleware
 from starlette.requests import Request
 from starlette.types import ASGIApp
@@ -40,4 +45,25 @@ class CacheControlMiddleware(BaseHTTPMiddleware):
             if request.method in ["HEAD", "GET"] and response.status_code < 500:
                 response.headers["Cache-Control"] = self.cachecontrol
 
+        return response
+
+
+class CatalogUpdateMiddleware(BaseHTTPMiddleware):
+    """Middleware to update the catalog cache."""
+
+    async def dispatch(self, request: Request, call_next):
+        """Fetch Catalog either immediately or in background."""
+        if request.query_params.get("refresh"):
+            logger.debug("Forcing catalog refresh.")
+            await register_collection_catalog(request.app)
+            response = await call_next(request)
+        elif datetime.now() > request.app.state.catalog_expire:
+            logger.debug("Running catalog refresh in background.")
+            response = await call_next(request)
+            response.background = BackgroundTask(
+                register_collection_catalog, request.app
+            )
+        else:
+            logger.debug("No need to refresh catalog.")
+            response = await call_next(request)
         return response
--- a/tipg/middleware.py
+++ b/tipg/middleware.py
@@ -1,8 +1,8 @@
 """tipg middlewares."""
 
 import re
-from datetime import datetime
-from typing import Optional, Set
+from datetime import datetime, timedelta
+from typing import Any, Optional, Set
 
 from tipg.db import register_collection_catalog
 from tipg.logger import logger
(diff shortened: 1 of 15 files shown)


tipg/middleware.py
```diff
@@ -41,3 +46,24 @@ async def dispatch(self, request: Request, call_next):
...
+    async def dispatch(self, request: Request, call_next):
+        """Fetch Catalog either immediately or in background."""
+        if request.query_params.get("refresh"):
+            logger.debug("Forcing catalog refresh.")
+            await register_collection_catalog(request.app)
+            response = await call_next(request)
+        elif datetime.now() > request.app.state.catalog_expire:
+            logger.debug("Running catalog refresh in background.")
+            response = await call_next(request)
+            response.background = BackgroundTask(
```
ranchodeluxe, line 63: One of my favorite things in programming is when folks choose words that obscure how it actually works 😆 . A `BackgroundTask` is just a callback [AFAICS](https://github.com/encode/starlette/blob/4bab981d9e870f6cee1bd4cd59b87ddaf355b2dc/starlette/responses.py#L173-L174).

ranchodeluxe (approved):
I like this approach, nice work 👍 

Just one potential question. It's my understanding that `register_collection_catalog` actually takes things offline for a second. I might have that wrong. But if I don't then this change could cause even more 5xx(s). Thoughts?

tipg/settings.py
```diff
@@ -175,6 +176,8 @@ class DatabaseSettings(pydantic.BaseSettings):
 
     only_spatial_tables: bool = True
 
+    catalog_ttl: timedelta = timedelta(minutes=5)
```
vincentsarago, line 179: What do you think about moving this to the `APISettings` ?

tipg/middleware.py
```diff
@@ -41,3 +46,24 @@ async def dispatch(self, request: Request, call_next):
...
+
+class CatalogUpdateMiddleware(BaseHTTPMiddleware):
+    """Middleware to update the catalog cache."""
+
+    async def dispatch(self, request: Request, call_next):
+        """Fetch Catalog either immediately or in background."""
+        if request.query_params.get("refresh"):
+            logger.debug("Forcing catalog refresh.")
+            await register_collection_catalog(request.app)
+            response = await call_next(request)
```
vincentsarago, line 59: I don't think we need to set a special case for this. IMO the `refresh` will update the `catalog_expire` so it should not reach the next if block

bitner: The /refresh endpoint is only activated if settings.debug is TRUE. 

With this middleware there is no way to no for sure that you are getting a non-cached catalog without having this option since normally, the refresh is only ever happening in a background task and there is no way to know that it has run, while with this option it will block processing the request until it has refreshed the catalog if you want to make sure that you have the latest catalog.

vincentsarago: The issue with passing a `refresh` query parameter is that then you have to make sure any of the application endpoint won't need or refuse such keyword 🤷 (e.g: https://github.com/developmentseed/tipg/blob/main/tipg/dependencies.py#L223-L249)

vincentsarago: also imagine if someone pass `?refresh=False` it will refresh the catalog because the `False` will be passed as string not as bool

vincentsarago: > With this middleware there is no way to know for sure that you are getting a non-cached catalog

IMO, that's not a big deal! we could add a `last_updated` info in the `/collections` endpoint if this make things easier

bitner: All fair points. We can get rid of it. I might still want to add that functionality back in later -- as I think about it, it may be better to check the headers for Cache-Control=no-cache to trigger something like this.

vincentsarago: a custom `X-header` would be better. 

I'm about to submit a PR that change some of the logic.

ranchodeluxe:
> I'm naive here and might have this wrong -- but there's still a race condition here. While the background task is refreshing the catalog all concurrent requests will fail and 5xx right? I'm over here thinking of ways to make the refresh logic out-of-band with the request/response lifecycle during blue/green deploys (k8s and ECS) 🤔

If I have ☝️ this wrong then isn't  a `BackgroundTask` actually blocking the event loop? And doesn't that mean on large catalogs we'll have requests timing out for this period?

ranchodeluxe:
> > I'm naive here and might have this wrong -- but there's still a race condition here. While the background task is refreshing the catalog all concurrent requests will fail and 5xx right? I'm over here thinking of ways to make the refresh logic out-of-band with the request/response lifecycle during blue/green deploys (k8s and ECS) 🤔
> 
> If I have ☝️ this wrong then isn't a `BackgroundTask` actually blocking the event loop? And doesn't that mean on large catalogs we'll have requests timing out for this period?

Give the way we are using `BackgroundTask` here and based on [the source code](https://github.com/encode/starlette/blob/master/starlette/background.py#L25-L28) we are not blocking the event loop b/c it's running in a separate thread pool but that does (at least for large catalogs) open us up to my first assumption about race conditions i believe

ranchodeluxe:
> @ranchodeluxe as long as we are using a static catalog_collection we cannot be 💯 sure that a request will fetch a refreshed database.

That's not the guarantee I'm really looking for. Are you both sure this isn't gonna cause more 5xx's?

Also, what's the opt out option for this?

vincentsarago:
> Also, what's the opt out option for this?

`TIPG_TTL_CATALOG=0`

I think we should set `TTL` to 0 anyway

PR #70, opened 2023-05-09
Refactor #68: Database Catalog TTL

vincentsarago:
This PR updates #68:

- move settings to the API settings 
- move state variable to the `Database` object (`last_updated`) 
- ~~add `lastUpdated` in the `/collections` response (NOT IN THE SPECIFICATION)~~
- remove the `refresh` catch in the middleware

### DIFF
--- a/tipg/factory.py
+++ b/tipg/factory.py
@@ -25,7 +25,7 @@ from morecantile.defaults import TileMatrixSets
 from pygeofilter.ast import AstType
 
 from tipg import model
-from tipg.dbmodel import Collection
+from tipg.dbmodel import Collection, Database
 from tipg.dependencies import (
     CollectionParams,
     ItemsOutputType,
@@ -46,7 +46,7 @@ from tipg.resources.enums import MediaType
 from tipg.resources.response import GeoJSONResponse, SchemaJSONResponse
 from tipg.settings import FeaturesSettings, MVTSettings, TMSSettings
 
-from fastapi import APIRouter, Depends, Path, Query
+from fastapi import APIRouter, Depends, HTTPException, Path, Query
 from fastapi.responses import ORJSONResponse
 
 from starlette.datastructures import QueryParams
@@ -445,8 +445,13 @@ class OGCFeaturesFactory(EndpointsFactory):
             ),
         ):
             """List of collections."""
-            collection_catalog = getattr(request.app.state, "collection_catalog", {})
-            collections_list = list(collection_catalog.values())
+            collection_catalog: Database = request.app.state.collection_catalog
+            if not collection_catalog:
+                raise HTTPException(
+                    status_code=500, detail="Could not find collections."
+                )
+
+            collections_list = list(collection_catalog["collections"].values())
 
             limit = limit or 0
             offset = offset or 0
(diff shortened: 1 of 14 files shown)


tipg/settings.py
```diff
@@ -104,6 +103,8 @@ class APISettings(pydantic.BaseSettings):
 
     add_tiles_viewer: bool = True
 
+    catalog_ttl: int = 300
```
vincentsarago, line 106: Change the `catalog_ttl` to be a string so we can easily set it with env variable 

```
TIPG_CATALOG_TTL=10 uvicorn tipg.main:app --port 8080 --reload
```

tipg/middleware.py
```diff
@@ -51,19 +51,38 @@ async def dispatch(self, request: Request, call_next):
...
-            await register_collection_catalog(request.app)
-            response = await call_next(request)
-        elif datetime.now() > request.app.state.catalog_expire:
+        collection_catalog = getattr(request.app.state, "collection_catalog", {})
+        collections = collection_catalog.get("collections")
+        last_updated = collection_catalog.get("last_updated")
+
+        if not collections or datetime.now() > (
+            last_updated + timedelta(seconds=self.ttl)
+        ):
```
vincentsarago, line 77: 2 cases:
- there is not collections: we update (🤷‍♂️) 
- collections are too old: we update

tipg/middleware.py
```diff
@@ -51,19 +51,38 @@ async def dispatch(self, request: Request, call_next):
 class CatalogUpdateMiddleware(BaseHTTPMiddleware):
     """Middleware to update the catalog cache."""
 
+    def __init__(
+        self,
+        app: ASGIApp,
+        ttl: int = 300,
```
vincentsarago, line 57: we could maybe add a special case when `ttl=0` meaning that we don't want any refresh 🤷‍♂️

vincentsarago: that could be handled at the application level
```python
if settings.catalog_ttl:
    app.add_middleware(CatalogUpdateMiddleware, ttl=settings.catalog_ttl)

**Comment by vincentsarago** [comment]
tests are failing for some weird rounding issue in python 3.11. Not related to any change in this PR

File: tipg/factory.py [hunk 4]
```diff
@@ -527,6 +532,7 @@ def collections(
                 links=links,
                 numberMatched=matched_items,
                 numberReturned=items_returned,
+                lastUpdated=str(collection_catalog["last_updated"]),
```
bitner, line 535: let's just leave this out since it's not part of the spec

vincentsarago: 👍

tipg/middleware.py
```diff
@@ -51,19 +51,38 @@ async def dispatch(self, request: Request, call_next):
 class CatalogUpdateMiddleware(BaseHTTPMiddleware):
```
bitner, line 51: I thought the biggest thing you wanted to do was to get rid of BaseHTTPMiddleware???

vincentsarago: I will in an other PR

bitner (approved):
I'm cool with all these changes. I thought the biggest thing you wanted to do though was to get rid of BaseHTTPMiddleware though???

tipg/middleware.py
vincentsarago: changed this to only rely on the `last_updated` tag. 

Starting the application with a brand new database could in theory mean that every request would try to refresh the catalog, which IMO seems a bit risky

vincentsarago: yeah but what if the `startup` fails or if there is not collections at all 🤷 

I think it's fine as is right now!

Issue #45, opened 2023-03-22
Links inference 


vincentsarago:
So with #32 we split the OGC Features and Tiles Factories, in the `/` landing pages we are still getting links for Features and Tiles services when using the unified `Endpoints` factories but within each **collections** links we don't have links to the `Tiles` endpoints (because we have no easy ways to know if there is a tiles/ endpoint within the same route or application (well we could find some way but we'll prefer a more `general` solution).

I think the idea is to be able to infer any links within Collection and/or Items, by maybe following something like what Jeff did in https://github.com/stac-utils/stac-fastapi/pull/472

PR #83, opened 2023-06-30
[TO BE DISCUSSED] refactor factories to enable more customization

vincentsarago:
resolves https://github.com/developmentseed/tipg/issues/45

This PR does:
- refactor the main `Endpoints` Factory to inherit from both Features and Tiles factories (so we can reuse methods)
- export links to separate methods (features/tiles/common links)
- create a new Catalog dependency
- add module variables for conformance classes

### DIFF
--- a/tipg/main.py
+++ b/tipg/main.py
@@ -85,7 +85,7 @@ templates = Jinja2Templates(
 ogc_api = Endpoints(
     title=settings.name,
     templates=templates,
-    with_tiles_viewer=settings.add_tiles_viewer,
+    with_viewer=settings.add_tiles_viewer,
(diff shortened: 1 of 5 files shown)


tipg/main.py
```diff
@@ -85,7 +85,7 @@ async def lifespan(app: FastAPI):
 ogc_api = Endpoints(
     title=settings.name,
     templates=templates,
-    with_tiles_viewer=settings.add_tiles_viewer,
+    with_viewer=settings.add_tiles_viewer,
```
vincentsarago, line 88: need to be added in the changelog!

PR #86, opened 2023-07-05
fix incorrect import in example

jackharrhy:
example uses PostgresSettings, but imports DatabaseSettings, classes do not seem compatible so this change makes the import match what is used in the example


vincentsarago:
🙏 thanks @jackharrhy

PR #87, opened 2023-07-07
Small fix to transform to 4326

RemcoMeeuwissen:
## What I am changing

We found a small bug where, if the geom is not 4326 it should transform to 4326 but instead the transform uses the SRID which the geom is already in.

Should [line 97](https://github.com/RemcoMeeuwissen/tipg/blob/main/tipg/sql/dbcatalog.sql#L97) also be changed? Because it's also going to use the geom native SRID rather than 4326.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -10,36 +10,37 @@ Note: Minor version `0.X.0` update might break the API, It's recommended to pin 
 
 ### Added
 
-* `type` query parameter to filter collections based on their type (`Function` or `Table`)
+- `type` query parameter to filter collections based on their type (`Function` or `Table`)
+- fixed a small bug in the `tipg_properties` SQL function where the bounds property was not properly transformed to 4326
 
 ## [0.2.0] - 2023-06-22
 
 ### Changed
 
-* rename `tipg.db` -> `tipg.database`
-* rename `tipg.dbmodel` -> `tipg.collections`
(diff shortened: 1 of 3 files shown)


vincentsarago:
🙏 it would be nice if we could test this

RemcoMeeuwissen:
Took a few days but this test should be enough, it at least covers the use case where I found the bug :slightly_smiling_face:

vincentsarago:
🙏 @RemcoMeeuwissen could you add a line in the changelog please 

Thank you

RemcoMeeuwissen:
Didn't realize that there is a changelog now, hope this is clear enough

CHANGES.md
```diff
@@ -10,36 +10,37 @@ Note: Minor version `0.X.0` update might break the API, It's recommended to pin
 
 ### Added
 
-* `type` query parameter to filter collections based on their type (`Function` or `Table`)
+- `type` query parameter to filter collections based on their type (`Function` or `Table`)
+- fixed a small bug in the `tipg_properties` SQL function where the bounds property was not properly transformed to 4326
```
vincentsarago, line 14:
```suggestion
- fixed a small bug in the `tipg_properties` SQL function where the bounds property was not properly transformed to 4326  (author @RemcoMeeuwissen, https://github.com/developmentseed/tipg/pull/87)
```

vincentsarago (approved):

PR #89, opened 2023-07-13
Fixed a bug where floats in the database would not be serialized corr…

RemcoMeeuwissen:
## What I am changing

 Ran into a bug where having a float in the database would cause a `Type is not JSON serializable: decimal.Decimal` error when requesting a json response.

## How I did it

orjson has a default parameter when calling the dumps function which allows you to specify how to serialize unknown types, using this Decimal is now serialized to string.

Because the ORJsonResponse wrapper provided by fastapi does not support sending along this parameter I tweaked the GeoJSONResponse and SchemaJSONResponse classes to use orjson directly (basically recreating the wrapper, luckily it's fairly small).

## How you can test it

I've added a number field to the my_data.sql fixture, this caused the bug to be triggered on existing tests so no new tests are needed.


tipg/resources/response.py
```diff
@@ -1,15 +1,42 @@
...
 
     media_type = "application/geo+json"
 
+    def render(self, content: Any) -> bytes:
+        """Render the content into a JSON response using orjson"""
+        return orjson.dumps(
+            content,
+            default=default,
+            option=orjson.OPT_NON_STR_KEYS | orjson.OPT_SERIALIZE_NUMPY,
+        )
```
vincentsarago, line 28: could we create a custom JSONResponse class to avoid duplicating the `render` code?

RemcoMeeuwissen: Yeah should be easy

RemcoMeeuwissen: Done :slightly_smiling_face:

vincentsarago:
@RemcoMeeuwissen I've been away and focused on titiler for the past week, I'll try to have a loot on all the contributions next week 🙏

tipg/resources/response.py
bitner: Looks good to me. Can you check if we need to do the same thing for postgres fields of type "numeric" as well (which is equivalent to the postgres "decimal" type?

tipg/resources/response.py
```diff
@@ -1,6 +1,29 @@
...
+
+import orjson
+
+from fastapi.responses import JSONResponse
+
+
+def default(obj):
+    """Instruct orjson what to do with types it does not natively serialize"""
+    if isinstance(obj, decimal.Decimal):
+        return str(obj)
```
vincentsarago, line 14: why do we use `string` instead of `float` here? 

The result of this will return a string within the JSON so we loose the fact that it is a number in the table

cc @RemcoMeeuwissen @bitner

tests/fixtures/my_data.sql
```diff
@@ -7,12 +7,14 @@ SELECT AddGeometryColumn('public','my_data','geom',4326,'GEOMETRY',2);
...
-INSERT INTO "public"."my_data" ("geom" , "id", "datetime") VALUES ('0103000020E6100000010000000D000000BBE9944235C347C0EBF06E7961EE52406ADE718A8EC447C0D122DBF97EEE5240942D6301ECB947C05B59871F60F0524086CAEEF61AAE47C0BDEF3BBB76F252400A4C8422590E46C0B656FB86F03B5340FA28B2217F0346C0CE0A257ADB3D534057EC2FBB27F745C02B1895D409405340BEE6287052F545C01AA33BF2DF3F53401D386744692743C07958A835CDFF52403EDF52396E3443C0DC9EAF2DC7FD5240B9E39237FD0645C0574B4E2543B552400AD7A3703D1245C03A234A7B83B35240BBE9944235C347C0EBF06E7961EE5240', '5', '2004-10-24 10:23:54');
+ALTER TABLE "public"."my_data" ADD COLUMN "decimal" DECIMAL;
+ALTER TABLE "public"."my_data" ADD COLUMN "numeric" NUMERIC;
```
vincentsarago, line 11: I have added values for `numeric` and it seems that they get returned as `Decimal` so we're fine

tests/routes/test_items.py
```diff
@@ -769,6 +769,10 @@ def test_items_env_table_config_main(app, monkeypatch):
     body = response.json()
     assert body["features"][0]["geometry"]["type"] == "Polygon"
 
+    # Make sure that Postgres Decimal and Numeric are converted to str
+    assert isinstance(body["features"][0]["properties"]["decimal"], str)
+    assert isinstance(body["features"][0]["properties"]["numeric"], str)
```
vincentsarago, line 774: TBD ☝️

tipg/resources/response.py
bitner: I'm not actually sure what we want here as best practice as decimal/numeric can be arbitrarily larger than what float could handle

RemcoMeeuwissen: Yeah string felt like the safe option to me so that's why I went with it, but I don't know at which point precision is lost so I don't know if it is actually a relevant concern.

bitner: I'd say that we stick with string as it is the only way we can be certain to not be lossy. If someone knows their data fits into a float, they can alter their table.

bitner (approved):
I think this is good to go.

tipg/resources/response.py
vincentsarago: Looking at `simplejson` library, it seems they do convert decimal to float 
```python
a = {"a": decimal.Decimal('1.00000000006')}

simplejson.dumps(a)
'{"a": 1.00000000006}'
```

vincentsarago: 🤔 I think it's because, simplejson will automatically convert decimal to float internally! 

> I'd say that we stick with string as it is the only way we can be certain to not be lossy. If someone knows their data fits into a float, they can alter their table.

👍 let's merge this, and we will revisit later if needed 🙏

CHANGES.md
```diff
@@ -14,6 +14,7 @@ Note: Minor version `0.X.0` update might break the API, It's recommended to pin
 - `type` query parameter to filter collections based on their type (`Function` or `Table`)
 - fixed a small bug in the `tipg_properties` SQL function where the bounds property was not properly transformed to 4326 (author @RemcoMeeuwissen, https://github.com/developmentseed/tipg/pull/87)
 - handling functions that are interpreted as collections but lack parameters (author @jackharrhy, https://github.com/developmentseed/tipg/pull/96)
+- fixed a bug where orjson could not properly serialize float values present in the properties of a feature (author @RemcoMeeuwissen, https://github.com/developmentseed/tipg/pull/89)
```
vincentsarago, line 17: I'll update this line to mention that Decimal will be converted to string

PR #94, opened 2023-07-17
generate popup contents without string concat.

jackharrhy:
## What I am changing

- Previous version of popup content generation is vulnerable to [XSS attacks](https://owasp.org/www-community/attacks/xss/) due to crafting raw HTML from feature properties.

## How I did it

- Instead of generating the popup contents from string concatenation, generate from using document.`...` APIs instead
- This method of creating content, while more verbose, should not be vulnerable to XSS due to any feature content being added as innerText

## How you can test it

Insert this table into the db:

```sql
CREATE TABLE xss (name varchar primary key, geom geometry, description varchar);
INSERT INTO xss VALUES ('XSS Test', 'POINT(0 0)', '<p onclick="alert(1)">click me!</p>');
```

Navigate to http://localhost:8000/collections/public.xss/items/XSS%20Test

On current main: notice that clicking on the popup, and then clicking on the text, will popup with an alert:

image

On this branch, the popup will render the true contents, and is not clickable as before:

image

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,8 +11,8 @@ Note: Minor version `0.X.0` update might break the API, It's recommended to pin 
 ### Added
 
 - `type` query parameter to filter collections based on their type (`Function` or `Table`)
-- fixed a small bug in the `tipg_properties` SQL function where the bounds property was not properly transformed to 4326  (author @RemcoMeeuwissen, https://github.com/developmentseed/tipg/pull/87)
-- added popups to leaflet maps on `items` and `item` page. (author @krishnaglodha, https://github.com/developmentseed/tipg/pull/91)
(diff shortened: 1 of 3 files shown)


jackharrhy:
fyi @krishnaglodha

krishnaglodha:
Ohh I see. This one makes more sense. Thanks a lot for clarification

vincentsarago:
🥳 thanks @jackharrhy 
could you update https://github.com/developmentseed/tipg/blob/main/CHANGES.md?plain=1#L15 to add link to this PR and your name as co-author of this feature 🙏

jackharrhy:
done @vincentsarago

vincentsarago:
🥳

PR #96, opened 2023-07-19
fix function based collections without params

jackharrhy:
## What I am changing

- Currently, parameter-less functions are broken on startup when the Collection model is instantiated by them

## How I did it

- Handle the table dict having an explicit `parameters: None` properly, always default to a list to make pydantic happy

## How you can test it

Create the following function
```sql
CREATE OR REPLACE FUNCTION public.test() RETURNS TABLE(foo integer, location geometry)
AS 'SELECT 1, ST_MakePoint(0,0);' LANGUAGE SQL;
```

### Currently

Launching tipg will fail when it goes to create an instance of Collection based off of this function due to a validation error

### After this PR

Launching tipg will work as expected without any validation errors

### DIFF
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -970,6 +970,9 @@ async def get_collection_index(  # noqa: C901
                     ):
                         geometry_column = c
 
+            if table.get("parameters") is None:
+                table["parameters"] = []
+
             catalog[table_id] = Collection(
                 type=table["entity"],
                 id=table_id,
@@ -980,7 +983,7 @@ async def get_collection_index(  # noqa: C901
                 properties=properties,
                 datetime_column=datetime_column,
                 geometry_column=geometry_column,
-                parameters=table.get("parameters", []),
+                parameters=table["parameters"],
             )
 
         return Catalog(collections=catalog, last_updated=datetime.datetime.now())


jackharrhy:
@vincentsarago if you'd like a test for this let me know and i can do that as well :)

vincentsarago:
Thanks @jackharrhy 

> if you'd like a test for this let me know and i can do that as well :)

Yes please 🙏

bitner:
I might be missing something, but I don't see how this is doing anything differently than `parameters=table.get("parameters", [])`?

vincentsarago:
@bitner because I guess `table` can have `parameters=None`.

Maybe this tell us there is a bug and that parameters should always default to `[]` or not be set in table if empty

vincentsarago:
@bitner 

>Maybe this tell us there is a bug and that parameters should always default to [] or not be set in table if empty

This is what `table` look like
```
{'name': 'test', 'entity': 'Function', 'schema': 'public', 'parameters': None, 'properties': [{'name': 'foo', 'type': 'integer'}, {'name': 'location', 'type': 'geometry', 'geometry_type': 'Geometry'}]}
```

You can see that `tipg_catalog` function returns `None` for `parameters`

tipg/collections.py
```diff
@@ -980,7 +983,7 @@ async def get_collection_index(  # noqa: C901
                 properties=properties,
                 datetime_column=datetime_column,
                 geometry_column=geometry_column,
-                parameters=table.get("parameters", []),
+                parameters=table["parameters"],
```
vincentsarago, line 986: it might be easier just to do `parameters=table.get("parameters") or [],`

vincentsarago:
I have updated the code and added tests. @bitner I think we don't need to change the `tipg_catalog` SQL code to return `[]` by default for now!

jackharrhy:
Thanks for taking this over and getting it sorted!

PR #102, opened 2023-07-28
update to pydantic 2.0

vincentsarago:


tests/routes/test_item.py
```diff
@@ -10,6 +12,7 @@ def test_item(app):
     assert body["type"] == "Feature"
     assert body["id"] == 1
     assert body["links"]
+    Item.model_validate(body)
```
vincentsarago, line 15: because we don't use `pydantic model` within the `/items` and `/items/{id}` endpoints it's good to `validate` the responses in tests

PR #108, opened 2023-08-04
fix custom sql dir not optional

jackharrhy:
- set custom_sql_directory to default to None
- updated changelog

---

Without this, if no `custom_sql_directory` is specified, starting the default tipg.main:app will fail with a pydantic error


vincentsarago (approved):

vincentsarago:
🤦 thanks for the fix

jackharrhy:
@vincentsarago this one always catches me out myself! :laughing: , be nice of optional would be default to None automagically, but alas...

PR #113, opened 2023-08-25
forward dependency to Factories

vincentsarago:


vincentsarago:
Note:

I just realized that we are using `CollectionParams` in `properties_filter_query` and `function_parameters_query` so if a user use a custom `CollectionParams` those dependency might not work!

PR #116, opened 2023-08-26
WIP: CatalogId Params with --root-path Prefix

ranchodeluxe:
Just a draft for discussion of using `{catalogId}` as a param in the path.

k8s: https://github.com/developmentseed/eoapi-k8s/blob/a7ca03105aeed425e787b7315c477b3c520c1f5e/helm-chart/eoapi/eoapi-manfests-catalogid-with-prefix.yml

### DIFF
--- a/tipg/settings.py
+++ b/tipg/settings.py
@@ -141,8 +141,8 @@ class PostgresSettings(BaseSettings):
             username=info.data.get("postgres_user"),
             password=info.data.get("postgres_pass"),
             host=info.data.get("postgres_host", ""),
-            port=info.data.get("postgres_port", 5432),
-            path=f"/{info.data.get('postgres_dbname') or ''}",
+            port=int(info.data.get("postgres_port", 5432)),
+            path=f"{info.data.get('postgres_dbname') or ''}",
         )
 
 
@@ -179,5 +179,5 @@ class CustomSQLSettings(BaseSettings):
 
 
 class HostToSchemaLookupSettings(BaseSettings):
-    mapping: Dict[str, List[str]] = {}
+    mapping: Dict[str, Optional[Dict[str, List[str]]]] = {}
     model_config = {"env_prefix": "TIPG_MULTI_", "env_file": ".env", "extra": "ignore"}
--- a/tipg/settings.py
+++ b/tipg/settings.py
@@ -176,3 +176,8 @@ class CustomSQLSettings(BaseSettings):
             return list(self.custom_sql_directory.glob("*.sql"))
 
         return None
+
+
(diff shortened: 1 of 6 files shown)


ranchodeluxe:
related to: https://github.com/NASA-IMPACT/veda-data/issues/19

tipg/settings.py
```diff
@@ -141,8 +141,8 @@ def assemble_db_connection(cls, v: Optional[str], info: FieldValidationInfo) ->
             username=info.data.get("postgres_user"),
             password=info.data.get("postgres_pass"),
             host=info.data.get("postgres_host", ""),
-            port=info.data.get("postgres_port", 5432),
-            path=f"/{info.data.get('postgres_dbname') or ''}",
+            port=int(info.data.get("postgres_port", 5432)),
```
vincentsarago, line 144: oh that's definitely a bug I believe

ranchodeluxe: yes indeedy

vincentsarago:
@ranchodeluxe have you seen https://github.com/developmentseed/tipg/discussions/118 ?

ranchodeluxe:
> @ranchodeluxe so you want `{catalog_name}.hostname.io/collections/{schema}.{table}`?
> 

1. I think I'd rather just use the hostname as the new catalog name b/c I don't want to make assumptions about the domain structxure as it could be subdomain (xyz.<domain>.com) or the raw domain (yadda.com). Let me look deeper into how to do this

2. Regardless of which route we go with this I think you probably need to make the high level decision about whether having multiple catalogs is something you see `tipg` supporting natively through configuration or only through examples and extension. Because somewhere there has to be catalog-name-to-schema-name mapping unless everyone is hard coding it like your examples above show

vincentsarago:
> Regardless of which route we go with this I think you probably need to make the high level decision about whether having multiple catalogs is something you see tipg supporting natively through configuration or only through examples and extension. Because somewhere there has to be a catalog-name-to-schema-name mapping unless everyone is just extending through examples and hard coding it like your examples above shows in async def lifespan

At this point supporting multiple catalog seems possible using custom code (as shown in https://github.com/developmentseed/tipg/pull/116#issuecomment-1696132905 or https://github.com/developmentseed/tipg/discussions/118). Because it's not (yet) part of the OGC API I don't think this is something we would want to `officially` support. We choose to support multiple schemas (mixed together) as one catalog but again as shown, this can be customized. If at one point the OGC Features API adds a `catalog` part we may revisit this to add `/catalogs/{catalog_id}` with the `catalog_id` being the schema

I think what you are discussing is really an infrastructure issue.

PR #119, opened 2023-08-27
fix pg settings type and add tests

vincentsarago:


tipg/settings.py
```diff
@@ -131,18 +131,20 @@ class PostgresSettings(BaseSettings):
...
         if isinstance(v, str):
-            return v
+            return PostgresDsn(v)
 
         return PostgresDsn.build(
             scheme="postgresql",
             username=info.data.get("postgres_user"),
             password=info.data.get("postgres_pass"),
             host=info.data.get("postgres_host", ""),
             port=info.data.get("postgres_port", 5432),
```
vincentsarago, line 146: pydantic will always cast the host to `integer` so there is no real need to add `int()` here, even if it would be technically correct.

PR #120, opened 2023-09-05
use older function to support old Postgres

### LINKED ISSUE #67: `string_to_table` fails with Postgres < 14
```
asyncpg.exceptions.UndefinedFunctionError: function string_to_table(text, unknown) does not exist
HINT:  No function matches the given name and argument types. You might need to add explicit type casts.
```

https://github.com/developmentseed/tipg/blob/main/tipg/sql/dbcatalog.sql#L131-L139

Result of running `uvicorn tipg.main:app`

vincentsarago:
closes #67 

This PR replace `string_to_table(...)` by `unnest(string_to_array(...))` to support older postgres 

> string_to_table() provides the set-based equivalent to [string_to_array()](https://pgpedia.info/s/string_to_array.html), and is functionally the same albeit somewhat faster than [unnest](https://pgpedia.info/u/unnest.html)(string_to_array())

From PG docs: https://pgpedia.info/s/string_to_table.html

Note: we could also use `regexp_split_to_table` https://pgpedia.info/r/regexp_split_to_table.html which seems to give the same result 🤷

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,12 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/).
 
 Note: Minor version `0.X.0` update might break the API, It's recommended to pin `tipg` to minor version: `tipg>=0.1,<0.2`
 
+## [0.4.4] - TBD
+
+### fixed
+
+- replace `string_to_table(...)` by `unnest(string_to_array(...))` to support Postgres<14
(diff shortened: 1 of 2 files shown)


vincentsarago:
👋 @bitner ☝️

bitner (approved):

PR #137, opened 2023-10-31
add items/item depencencies

vincentsarago:
This PR adds dependencies for both `/items` and `/items/{itemID}` endpoints 

This will enable custom backend which could allow more/less query parameters


vincentsarago:
🤔 I think this is over-engineered if user need to override endpoint input it will be much easier to re-write most of the endpoints like in https://github.com/developmentseed/tipgstac/blob/main/tipgstac/factory.py

PR #140, opened 2023-11-15
add fix for using is null

bitner:
Adds a fix and tests for a bug in handling is null in both cql2 text and cql2 json.


vincentsarago (approved):

vincentsarago:
thanks @bitner 

I'll make a new release with this fix today

PR #143, opened 2023-11-22
make calculating spatial/datetime extents optional

hrodmn:
As discussed in #139, Calculating the extent of spatial/datetime columns can be slow enough that serverless deployments of `tipg` are not possible.

Skipping the extent calculations is one way to make application startup almost instantaneous! With this change, the user can set `TIPG_DB_SPATIAL_EXTENT=False` and `TIPG_DB_DATETIME_EXTENT=False` to skip these calculations.

I am not sure if passing these arguments all the way through to `pg_temp.tipg_properties` is the best way to accomplish this, but I got it working. If there is a cleaner way that you can think of please let me know!

### DIFF
--- a/docs/src/user_guide/configuration.md
+++ b/docs/src/user_guide/configuration.md
@@ -50,6 +50,8 @@ class: `tipg.settings.DatabaseSettings`
 prefix: **`TIPG_DB_`**
 
 - **SCHEMAS** (list of string): Named schemas, `tipg` can look for `Tables` or `Functions`. Default is `["public"]`
+- **SPATIAL_EXTENT** (bool): Calculate spatial extent of records. Default is `True`.
+- **DATETIME_EXTENT** (bool): Calculate temporal extent of records. Default is `True`.
 
 #### `Tables`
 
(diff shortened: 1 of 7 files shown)


vincentsarago:
Approach sounds 👌 

Let's add some tests to validate 🙏

hrodmn:
> Approach sounds 👌

The one thing I was wondering is if we should skip spatial/temporal columns altogether instead of passing those variables through to that function, but it seems like we need other attributes from the spatial columns besides `bounds` so I did it this way.
 
> Let's add some tests to validate 🙏

I added a few tests!

vincentsarago:
> The one thing I was wondering is if we should skip spatial/temporal columns altogether instead of passing those variables through to that function, but it seems like we need other attributes from the spatial columns besides bounds so I did it this way.

I think it's fine to have them separate. Most people will have spatial index (hopefully) so them might want to avoid only the temporal calculation

vincentsarago (approved):

vincentsarago:
Let's wait for @bitner 👍 👎

PR #149, opened 2023-12-18
Refactor to use pgmini for SQL construction and pycql2 for CQL parsing

bitner:
### WORK IN PROGRESS
There are tons of print statements littered all over the place and other remnants from iterating to get tests to pass.

This PR refactors the building of SQL to use pgmini and pycql2.

## Reasons for pgmini
* pgmini has a cleaner syntax and requires far less use of 'escape hatches' with raw sql than buildpg.
* pgmini has built in options to be able to support both asyncpg and psycopg bind parameter syntaxes
* pgmini has easier hooks to be able to make stricter type casts while allowing more native use of postgres text input

## Reasons for pycql2
* pycql2 has a much smaller focused code base
* pycql2 has a better test suite that can be leveraged to ensure full support of CQL2 syntax - these tests also provide examples of many different queries that should be supported
* pycql2 has fuller support of the CQL2 text and json specification
* pycql2 Lark much more closely derives from the actual BNF specification for CQL2 text

## Additional Enhancements in this PR
* Fixes numerous bugs with CQL2 expressions
* Support for Array Operators
* Support for all Temporal Operators
* Enhancements for spatial operators to make sure that any transforms are performed to constant parameters when dealing with data that is not in geographic coordinates
* Allows for the use of geography in addition to geometry which had very limited support previously

## TODO
- [ ] Update dependencies to remove buildpg / pygeofilter and add pycql2 and pgmini - this does require the non-merged branch of pycql2 with pydantic 2
- [ ] Remove all print statements
- [ ] Clean code up
- [ ] Figure out how to get past ruff/mypy errors with reuse of sql function in cql2sql.py using plum dispatch (current ci set up is bombing and won't allow use of combined noref/ignore comments to get rid of errors
- [ ] Add tests that compare CQL2 input to desired SQL output
- [ ] Add tests for array operators
- [ ] Add tests for temporal operators
- [ ] Add tests for geography


vincentsarago:
@bitner 👋 anything I can do to help here?

vincentsarago:
this can go now :-)

PR #160, opened 2024-01-09
Patch/do not exclude null properties

### LINKED ISSUE #159: Response MediaType CSV failing
Sometimes the CSV format download seems to fail. `f=geojson` does NOT fail in these same circumstances.

Here it works (4 rows of data):
https://api.cerulean.skytruth.org/collections/public.aoi_type/items
https://api.cerulean.skytruth.org/collections/public.aoi_type/items?f=csv

Here it fails (expect 141 rows of data, but returns 2 rows of garbled data):
https://api.cerulean.skytruth.org/collections/public.slick_plus/items?sortby=slick_timestamp&datetime=2023-12-01T00:00:00Z/2023-12-08T00:00:00Z&filter=machine_confidence%20GTE%200.6%20AND%20area%20GTE%2020000000
https://api.cerulean.skytruth.org/collections/public.slick_plus/items?sortby=slick_timestamp&datetime=2023-12-01T00:00:00Z/2023-12-08T00:00:00Z&filter=machine_confidence%20GTE%200.6%20AND%20area%20GTE%2020000000&f=csv

vincentsarago:
closes #159 

In the previous version (main) we were excluding `None` values, mostly to exclude empty geometries but sometimes it meant we were excluding valid properties, which could lead in `invalid` CSV headers (because row 1 and row 2 could have different properties, if row 1 had `None` property)

### DIFF
--- a/tipg/factory.py
+++ b/tipg/factory.py
@@ -754,21 +754,27 @@ class OGCFeaturesFactory(EndpointsFactory):
                 MediaType.json,
                 MediaType.ndjson,
             ):
-                rows = (
-                    {
-                        **{
-                            k: v
-                            for k, v in {
-                                "collectionId": collection.id,
-                                "itemId": f.get("id"),
-                                "geometry": f.get("geometry", None),
-                            }.items()
-                            if v is not None
-                        },
-                        **f.get("properties", {}),
-                    }
-                    for f in item_list["items"]
-                )
+                if any(
+                    [f.get("geometry", None) is not None for f in item_list["items"]]
+                ):
+                    rows = (
+                        {
+                            "collectionId": collection.id,
+                            "itemId": f.get("id"),
+                            **f.get("properties", {}),
+                            "geometry": f.get("geometry", None),
+                        }
+                        for f in item_list["items"]
+                    )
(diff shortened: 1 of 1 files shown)


tipg/factory.py
```diff
@@ -754,19 +754,27 @@ async def items(  # noqa: C901
                 MediaType.json,
                 MediaType.ndjson,
             ):
-                rows = (
-                    {
-                        k: v
-                        for k, v in {
+                if any(
+                    [f.get("geometry", None) is not None for f in item_list["items"]]
+                ):
```
vincentsarago, line 759: we first check if we have any data with `geometry`

tipg/factory.py
```diff
@@ -754,19 +754,27 @@ async def items(  # noqa: C901
...
+                    )
+                else:
+                    rows = (
+                        {
+                            "collectionId": collection.id,
+                            "itemId": f.get("id"),
+                            **f.get("properties", {}),
+                        }
+                        for f in item_list["items"]
+                    )
```
vincentsarago, line 777: here we can safely ignore `geometry` because of the previous test

PR #162, opened 2024-01-09
forward table columns in collection object

vincentsarago:
This PR fix the issue found in #161 

- add `table_columns` in the `Collection` object (non-filtered properties)
- change `id_column` type from `string` to `Column` object (so we don't need `id_column_info` method)

### DIFF
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -163,8 +163,9 @@ class Collection(BaseModel):
     dbschema: str = Field(..., alias="schema")
     title: Optional[str] = None
     description: Optional[str] = None
+    table_columns: List[Column] = []
     properties: List[Column] = []
-    id_column: Optional[str] = None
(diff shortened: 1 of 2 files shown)


tipg/collections.py
```diff
@@ -961,23 +951,27 @@ async def get_collection_index(  # noqa: C901
...
                     if p["name"] in fallback_key_names:
-                        id_column = p["name"]
+                        id_column = p
                         break
 
             datetime_column = None
             geometry_column = None
 
-            for c in properties:
+            for c in columns:
```
vincentsarago, line 974: ⚠️ we use the `full` columns instead of the filtered properties 

I feel this is a better way to do, if user want to change the geometry or datetime column we have to use the `geomcol` or `datetimecol` table config instead of filtering the properties

bitner (approved):

PR #187, opened 2024-08-27
Added in environment variable for sorting/config

mattdiez-at:
## Rationale

Column order is important for a few of the tables I work with and switching my users from their (database-defined) order to alphabetical caused a little shellshock. Essentially, getting this functionality will be the last part of our migration from a different feature server package.

## How it works

We add an environment variable `TIPG_SORT_COLUMNS` (which defaults to `True`). Set it to `False` to disable sorting of columns. It's essentially trivial code, plus accompanying docs and `Settings` model updates.

## Anything else?

Let me know if I need to add anything else here.

### DIFF
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -955,7 +955,7 @@ async def get_collection_index(  # noqa: C901
                 columns = sorted(table.get("properties", []), key=lambda d: d["name"])
             else:
                 columns = table.get("properties", [])
-            
+
             properties_setting = table_conf.properties or [c["name"] for c in columns]
 
             # ID Column
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -951,7 +951,11 @@ async def get_collection_index(  # noqa: C901
             table_conf = table_confs.get(confid, TableConfig())
 
             # Make sure that any properties set in conf exist in table
-            columns = sorted(table.get("properties", []), key=lambda d: d["name"])
+            if features_settings.sort_columns:
+                columns = sorted(table.get("properties", []), key=lambda d: d["name"])
+            else:
+                columns = table.get("properties", [])
+            
             properties_setting = table_conf.properties or [c["name"] for c in columns]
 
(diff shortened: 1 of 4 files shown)


tipg/settings.py
```diff
@@ -86,6 +86,7 @@ class FeaturesSettings(BaseSettings):
 
     default_features_limit: int = Field(10, ge=0)
     max_features_per_query: int = Field(10000, ge=0)
+    sort_columns: bool = True
```
vincentsarago, line 89: I wonder if we should have this set to `False` by default for the users that already use tipg 

I don't think it will hurt that much to be honest but 🤷 

Maybe we should sort the column in postgres directly 

cc @bitner

mattdiez-at: @vincentsarago - My sentiment was to do everything to leave current tipg behavior intact (I hate breaking stuff for established users), but if this were greenfield dev, I'd probably have turned sorting *off* by default.

tipg/collections.py
```diff
@@ -951,7 +951,11 @@ async def get_collection_index(  # noqa: C901
             table_conf = table_confs.get(confid, TableConfig())
 
             # Make sure that any properties set in conf exist in table
-            columns = sorted(table.get("properties", []), key=lambda d: d["name"])
+            if features_settings.sort_columns:
+                columns = sorted(table.get("properties", []), key=lambda d: d["name"])
+            else:
+                columns = table.get("properties", [])
```
vincentsarago, line 957:
```suggestion
            columns = table.get("properties", [])
            if features_settings.sort_columns:
                columns = sorted(columns, key=lambda d: d["name"])
```
let's save 1 line 😅

tipg/settings.py
vincentsarago, line 89: I think we should store this settings in the `TableSettings` class

mattdiez-at: Agreed.

PR #191, opened 2024-09-16
Adds the ability to use a schema other than pg_temp for installing catalog functions.

bitner:
Adds the tipg_schema setting to PostgresSettings defaulting to "pg_temp" to allow installing tipg catalog functions into a permanent schema (NOTE: this schema must already exist, and the logged in user must have full permissions to the schema!).

Basically, this is just doing a replace of "pg_temp" in the dbcatalog.sql with whatever is set in the tipg_schema setting.

### DIFF
--- a/tipg/database.py
+++ b/tipg/database.py
@@ -37,6 +37,7 @@ class connection_factory:
 
     async def __call__(self, conn: asyncpg.Connection):
         """Create connection."""
+        settings = PostgresSettings()
         await conn.set_type_codec(
             "json", encoder=orjson.dumps, decoder=orjson.loads, schema="pg_catalog"
         )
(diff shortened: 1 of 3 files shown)


tipg/settings.py
```diff
@@ -141,6 +141,8 @@ class PostgresSettings(BaseSettings):
 
     model_config = {"env_file": ".env", "extra": "ignore"}
 
+    tipg_schema: str = Field("pg_temp", pattern=r"[a-zA-Z][a-zA-Z0-9_-]*")
```
vincentsarago, line 144: - can you move this up, before the model_config 
- what do you think about `TIPG_SCHEMA_NAME` ?

tipg/database.py
```diff
@@ -61,10 +62,14 @@ async def __call__(self, conn: asyncpg.Connection):
 
         # Register custom SQL functions/table/views in pg_temp
         for sqlfile in self.user_sql_files:
-            await conn.execute(sqlfile.read_text())
+            await conn.execute(
+                sqlfile.read_text().replace("pg_temp", settings.tipg_schema)
```
vincentsarago, line 66: I'm trying to think about the use case for the custom function to have access the the `pg_temp` schema?

PR #199, opened 2024-12-12
Fallback to string for UUID columns

giorgiobasile:
Hi, I'm running into issues with UUID columns, basically being always returned as `null` in the OGC Features endpoints, whether they are recognized as feature `id` or be kept in the feature `properties`.

I pinpointed the issue to the default serialization function that handles only `Decimal` serialization to string, and added another case for `pgproto.UUID`, but wondering if there is a better way to handle non-json-serializable fields. Hope it helps!


vincentsarago:
Thanks for the PR @giorgiobasile 

would you able to provide a test case for this? 🙏

giorgiobasile:
Sure, just did!

giorgiobasile:
@vincentsarago I just updated `CHANGES.md`, is there any other change/action required on my end, or is it good to go?

PR #205, opened 2025-02-11
Fix CRS Transformation for custom TMS

callsumzg:
# Summary
This PR fixes issues in the spatial filtering logic for tile generation, which were causing missing features when custom TMS are used.

# Problem Description
When creating tiles with a custom TMS that has a different CRS (e.g. EPSG 2056), TiPG is constructing two different bounding boxes that aren’t in sync. One envelope is produced directly in the target CRS (e.g. 2056) while the other is produced in EPSG:4326 and then transformed into 2056. Because the transformation (and the segmentize step) may not be perfectly invertible, these two envelopes don’t match. As a result, only features that lie inside the intersection of both envelopes get selected. This effect is visible when zooming far enough into the dataset.

## Example:
When using a custom TMS for EPSG 2056 data (in this example, contour lines), the following SQL query is executed by TiPG for tile 11/1031/509:

```sql
WITH t AS (SELECT hoehe, in_kanton, t_id, ST_AsMVTGeom(ST_Transform(geom::geometry, 2056::int), ST_Segmentize(ST_MakeEnvelope(2683931.75, 1219444.29, 2684187.75, 1219700.29), 256.0), 4096, 256, True) AS geom
FROM hoehenkurven_1700.hk_10m
WHERE True AND ST_Intersects(ST_Transform(ST_Segmentize(ST_MakeEnvelope(8.54586926745684, 47.121968944662434, 8.549195345831953, 47.12430421660828, 4326), 0.003326078375112118), 2056::int), geom)
LIMIT 10000)
SELECT ST_AsMVT(t.*, 'hk_10m') FROM t
```

The two bboxes created by TiPG the two ST_MakeEnvelope functions are not congruent. 
They are off center, and only the contour lines that intersect with both bboxes are getting tiled by TiPG (see Screenshot below):

Bug

To fix the bug we must ensure to use one consistent envelope (bounding box) for both the geometry transformation for ST_AsMVTGeom and the spatial filter (ST_Intersects). 

We should compute the envelope once in one coordinate system. 
In TiPG, the envelope used for filtering is built in the _where() method (in collections.py) using a hard-coded SRID of 4326, whereas the envelope passed into ST_AsMVTGeom is already in (or transformed to) the target CRS (2056 in this example). As a result, only features that lie in the intersection of these two (misaligned) envelopes are returned.

With this fix, the SQL query above changes to this:
```sql
WITH t AS (SELECT hoehe, in_kanton, t_id, ST_AsMVTGeom(ST_Transform(geom::geometry, 2056::int), ST_Segmentize(ST_MakeEnvelope(2683931.75, 1219444.29, 2684187.75, 1219700.29), 256.0), 4096, 256, True) AS geom
FROM hoehenkurven_1700.hk_10m
WHERE True AND ST_Intersects(ST_Segmentize(ST_MakeEnvelope(2683931.75, 1219444.29, 2684187.75, 1219700.29, 2056), 256.0), geom)
LIMIT 10000)
SELECT ST_AsMVT(t.*, 'hk_10m') FROM t
```

# Changes Made
- In the _where() method:
    - Tile bounds are now directly unpacked and, if the geometry column’s SRID doesn’t match the TMS EPSG (defaulting to 4326 when needed), they’re transformed using pyproj.Transformer. This ensures spatial queries work correctly across different SRIDs, 
    - remove hardcoded EPSG 4326

# Impact
This commit fixes the tile generation so that tiles with data in different CRSs are rendered correctly and the tile response is robustly parsed.

### DIFF
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -12,6 +12,7 @@ from ciso8601 import parse_rfc3339
 from morecantile import Tile, TileMatrixSet
 from pydantic import BaseModel, Field, model_validator
 from pygeofilter.ast import AstType
+from pyproj import Transformer
 
 from tipg.errors import (
     InvalidDatetime,
@@ -540,8 +541,6 @@ class Collection(BaseModel):
             # Use a fallback of 4326 if tms.crs.to_epsg() returns a falsey value.
             tms_epsg = tms.crs.to_epsg() or 4326
             if geometry_column.srid != tms_epsg:
-                from pyproj import Transformer
-
                 transformer = Transformer.from_crs(
                     tms_epsg, geometry_column.srid, always_xy=True
                 )
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -532,31 +532,36 @@ class Collection(BaseModel):
             wheres.append(to_filter(cql, [p.name for p in self.properties]))
 
         if tile and tms and geometry_column:
-            # Get Tile Bounds in Geographic CRS (usually epsg:4326)
-            left, bottom, right, top = tms.bounds(tile)
+            # Get tile bounds in the TMS coordinate system
+            bbox = tms.xy_bounds(tile)
+            left, bottom, right, top = bbox
 
-            # Truncate bounds to the max TMS bbox
-            left, bottom = tms.truncate_lnglat(left, bottom)
-            right, top = tms.truncate_lnglat(right, top)
+            # If the geometry column’s SRID does not match the TMS CRS, transform the bounds:
+            # Use a fallback of 4326 if tms.crs.to_epsg() returns a falsey value.
+            tms_epsg = tms.crs.to_epsg() or 4326
+            if geometry_column.srid != tms_epsg:
+                from pyproj import Transformer
+
+                transformer = Transformer.from_crs(
+                    tms_epsg, geometry_column.srid, always_xy=True
+                )
+                left, bottom = transformer.transform(left, bottom)
+                right, top = transformer.transform(right, top)
 
             wheres.append(
                 logic.Func(
                     "ST_Intersects",
                     logic.Func(
-                        "ST_Transform",
+                        "ST_Segmentize",
                         logic.Func(
-                            "ST_Segmentize",
-                            logic.Func(
-                                "ST_MakeEnvelope",
-                                left,
-                                bottom,
-                                right,
-                                top,
-                                4326,
-                            ),
-                            right - left,
+                            "ST_MakeEnvelope",
+                            left,
+                            bottom,
+                            right,
+                            top,
+                            geometry_column.srid,
                         ),
-                        pg_funcs.cast(geometry_column.srid, "int"),
+                        right - left,
                     ),
                     logic.V(geometry_column.name),
                 )
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -544,8 +544,10 @@ class Collection(BaseModel):
                 transformer = Transformer.from_crs(
                     tms_epsg, geometry_column.srid, always_xy=True
                 )
-                left, bottom = transformer.transform(left, bottom)
-                right, top = transformer.transform(right, top)
+
+                left, bottom, right, top = transformer.transform_bounds(
+                    left, bottom, right, top
+                )
 
             wheres.append(
                 logic.Func(
--- a/tipg/collections.py
+++ b/tipg/collections.py
@@ -2,6 +2,7 @@
 
 import datetime
 import re
+from functools import lru_cache
 from typing import Any, Dict, List, Optional, Tuple, TypedDict, Union
 
 from buildpg import RawDangerous as raw
@@ -32,6 +33,8 @@ from fastapi import FastAPI
 
 mvt_settings = MVTSettings()
 features_settings = FeaturesSettings()
+
+TransformerFromCRS = lru_cache(Transformer.from_crs)
 
 
 def debug_query(q, *p):
@@ -541,7 +544,7 @@ class Collection(BaseModel):
             # Use a fallback of 4326 if tms.crs.to_epsg() returns a falsey value.
             tms_epsg = tms.crs.to_epsg() or 4326
             if geometry_column.srid != tms_epsg:
-                transformer = Transformer.from_crs(
+                transformer = TransformerFromCRS(
                     tms_epsg, geometry_column.srid, always_xy=True
                 )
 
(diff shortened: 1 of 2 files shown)


pascal-arnet-zg:
Looks promising

tipg/collections.py
```diff
@@ -532,31 +532,36 @@ def _where(  # noqa: C901
...
-            right, top = tms.truncate_lnglat(right, top)
+            # Get tile bounds in the TMS coordinate system
+            bbox = tms.xy_bounds(tile)
+            left, bottom, right, top = bbox
+
+            # If the geometry column’s SRID does not match the TMS CRS, transform the bounds:
+            # Use a fallback of 4326 if tms.crs.to_epsg() returns a falsey value.
+            tms_epsg = tms.crs.to_epsg() or 4326
+            if geometry_column.srid != tms_epsg:
+                from pyproj import Transformer
```
vincentsarago, line 543: let's put the `import` outside the if form

tipg/collections.py
```diff
@@ -532,31 +532,36 @@ def _where(  # noqa: C901
...
+
+                transformer = Transformer.from_crs(
+                    tms_epsg, geometry_column.srid, always_xy=True
+                )
+                left, bottom = transformer.transform(left, bottom)
+                right, top = transformer.transform(right, top)
```
vincentsarago, line 549: I usually use `transform_bounds` but not sure which is better to be honest

vincentsarago:
@callsumzg not sure if we could add a test for this change

callsumzg:
@vincentsarago What would you need for me to add a test for this? 
Data in EPSG 2056?

tipg/collections.py
callsumzg: Thanks for the feedback! I've moved the import outside the if statement. let me know if anything else needs adjustment..

tipg/collections.py
callsumzg: hm you are right, transform_bounds seems to be better here. according to the documentation it samples points along the edges and densifies the boundary, whereas transform() only transforms the corner points. i updated the code, thanks for the hint.

tipg/collections.py
```diff
@@ -532,31 +533,36 @@ def _where(  # noqa: C901
...
+            bbox = tms.xy_bounds(tile)
+            left, bottom, right, top = bbox
+
+            # If the geometry column’s SRID does not match the TMS CRS, transform the bounds:
+            # Use a fallback of 4326 if tms.crs.to_epsg() returns a falsey value.
+            tms_epsg = tms.crs.to_epsg() or 4326
+            if geometry_column.srid != tms_epsg:
+                transformer = Transformer.from_crs(
+                    tms_epsg, geometry_column.srid, always_xy=True
+                )
```
vincentsarago, line 546: what do you think about creating a LRU cache for transformers https://pyproj4.github.io/pyproj/stable/advanced_examples.html#caching-pyproj-objects

vincentsarago: > Therefore, my conclusion would be: You can implement it, but it probably won't make a difference.

Yeah it makes sense, I think most `transformer` would be from epsg:4326 to epsg:3857 which should be pretty fast to get from proj.

I'll re-think about this later. 

### Tile Caching 

As for the titiler project, we didn't include caching at the application level because it could complexify the  code base and add dependencies. We assumed that people would put their application behind a Cache layer like ` aws cloudfront`

callsumzg: ### Transformer Caching
i think transformer caching might make sense, but i'm more on the empirical side and would suggest performing A/B performance tests if you have experience with that or know someone who has?

### Tile Caching
I see your points about increased complexity and dependencies.

In our case, we have some vector datasets that update regularly (daily/weekly like official cadastre data) and others that update less often (like contour lines). In local tests with TiPG, I noticed that the initial tiling takes a very long time, especially with heavy database queries on many features , which is too slow.

For frequently updated datasets, I see two options:

1. No pre-tiling: First-time tile requests are slow.
2. Pre-tile common zoom levels (e.g., overnight): Cache them so subsequent requests are fast.

Since we run on-premises without AWS or Azure, local caching and pre-tiling are important for us. Just like we do with our Mapproxy-generated raster tiles :-)

**How do you handle slow initial tile requests after vector data updates? Wouldn't a mechanism to pre-tile or warm up the cache be beneficial?**

vincentsarago: > How do you handle slow initial tile requests after vector data updates? Wouldn't a mechanism to pre-tile or warm up the cache be beneficial?

I personally don't deal with that and we've been looking for more users feedback to see what kind of future development we could do 😄 

We should maybe try to ease the customization of the `Collection` class, so user could easily extent the class and implement caching on top of the `get_tile` method 



```

Comment (callsumzg) [reply tipg/collections.py:546]: > I personally don't deal with that and we've been looking for more users feedback to see what kind of future development we could do 😄

Now you have some 😆I can give you some more 😉
For us, the main reasons we don't serve vector tiles yet are:

- We need custom TMS support for our EPSG 2056 data, because our WebGIS uses Openlayers as a framework. Multiple TMS/CRS is not supported, nor are on the fly reprojections with vector tiles. Also, all our data is in EPSG 2056.
- pre-tiling, since our vector data updates quite often
- caching/storing the tiles for faster serving (yes, a cache server in front would take care of this, but also adds more complexity. And we use local file storage with Mapproxy for our raster tiles successfully for many years)
- OGC complience (❤️TiPG)
- Postgres-Support
- Good project maintenance and the possibility for contributions (❤️TiPG)



> We should maybe try to ease the customization of the Collection class, so user could easily extent the class and implement caching on top of the get_tile method

I’ll take another look at the changes I made for pre-tiling and caching. 
Perhaps you could give me some feedback, or it might inspire ideas for customizing the collection class?

Comment (vincentsarago) [reply tipg/collections.py:546]: Ok let's start another discussion about caching/customization outside this PR 🙏 

Going back at transform caching, I think we should do it 

```python
from functools import lru_cache
from pyproj import Transformer

TransformerFromCRS = lru_cache(Transformer.from_crs)

%timeit -n 5 TransformerFromCRS(2056, 4326, always_xy=True)
446 ns ± 224 ns per loop (mean ± std. dev. of 7 runs, 5 loops each)

%timeit -n 5 Transformer.from_crs(2056, 4326, always_xy=True)
1.59 ms ± 387 μs per loop (mean ± std. dev. of 7 runs, 5 loops each)
```

callsumzg: > Ok let's start another discussion about caching/customization outside this PR 🙏

I already started one in the Q&A https://github.com/developmentseed/tipg/discussions/201 👍 

> Going back at transform caching, I think we should do it
I'm in!

PR #207, opened 2025-02-22
add aiocache for `tile` cache

vincentsarago:


callsumzg:
> @callsumzg would love to get your opinions on this.
> 
> aiocache seems like the only `python async` lib of cache. Its seems extensible so users could provide its own cache backend or use one of the predefined classes

Finally had time to look into it. If i am honest, i'm not the biggest fan of relying only on one specific caching library. As far as i can tell, the draft also relies on a global config, which would make it not so easy to extend or modify later? But the main thing is that the library aiocache seems to be restricted to redis, memory and memcache, or not? all good solutions when you have not so much data and/or lots of memory/money :-) but as soon as you have more data and also want to pretile (which i am a big fan of, i would not implement on-demand caching for the most requested zoomlevels), you have some problems. I would rather implement a more modular approach with an abstract layer by defining a base cache interface and then implementig separate backend classes for file, redis, s3, sqlite/mbtiles, etc. This would have the benefits of decoupling the caching mechanism from the rest of the application and if you want to support a new caching backend (e.g geopackage, azureblob) it is as simple as adding a new implementation without changing the core. For example if we implemented several separate caching backends, one for file, one for redis, one for s3, and suddenly we have to replace the s3 library, we would only need to replace/update the specific code for the s3 and not as well for all the other cache-backends. :-) 

I dont know if i explained what i mean clearly. If you give me some time, i can put together a draft so you can see what i mean. 🙈😅

Btw, the [redis-py](https://github.com/redis/redis-py) library includes a built in [async support](https://redis.readthedocs.io/en/latest/examples/asyncio_examples.html) too

vincentsarago:
Quick note:

> But the main thing is that the library aiocache seems to be restricted to redis, memory and memcache, or not

You can write your own aiocache backend (e.g. ondisk), but by default it only provide those 3 backends.

> which would make it not so easy to extend or modify later

It wont if you were using `tipg` as it is but the idea of tipg is usually to be extended/customized to fit user needs

PR #218, opened 2025-05-16
feat: filter collection based on name

sanzog03:
## What I am changing

- Add the ability to filter collections based on collection ID.

## How I did it

- Add a query parameter depdency injection which checks for collection ID substring regex, validates the regex and uses that to filter the CollectionList. 

## How you can test it

- Using the API docs, check the list of collections without any filter. Then, select any collection and use part of the collection ID as input to the collectionID regex filter to check the result.
- A test is added to the code.

### DIFF
--- a/tests/routes/test_collections.py
+++ b/tests/routes/test_collections.py
@@ -372,3 +372,20 @@ def test_collections_temporal_extent_datetime_column(app):
     assert len(intervals) == 4
     assert intervals[0][0] == "2004-10-19T10:23:54+00:00"
     assert intervals[0][1] == "2007-10-24T00:00:00+00:00"
+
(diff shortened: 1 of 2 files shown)


slesaad:
@vincentsarago can you please have a look at this? we have this need for GHG center

vincentsarago:
Hi @sanzog03 @slesaad 

This feature would be outside OGC spec. I think it would make more sense to implement CQL filter like we have for `collections-search` in stac-api-spec https://github.com/stac-api-extensions/collection-search?tab=readme-ov-file#filter-cql

cc @hrodmn @bitner

vincentsarago:
Can we close this PR?

PR #219, opened 2025-05-16
feat: add extra property to collection and collection List

sanzog03:
## What I am changing

Add the ability to add extra parameters to collection and collectionList. Hence, they can be used to store metadata and extra information about the collection.

## How I did it

In the current implementation of TiPg, a schema table in represents the whole collection and the rows represent items.
Since there is no separate table to store the collection data, If we add any columns with prefix "collection_properties_", we use that as extra properties.
A dependency injection is added which handles the above mentioned case.

## How you can test it

First ingest data with a column prefixing with "collection_properties_". Then, using the API docs, test if it shows up in the collection and collection list endpoints.

### DIFF
--- a/tipg/dependencies.py
+++ b/tipg/dependencies.py
@@ -502,3 +502,94 @@ def CollectionsParams(
         next=offset + returned if matched - returned > offset else None,
         prev=max(offset - limit, 0) if offset else None,
     )
+
+ExtraProperties = Dict[str, any]
+
+async def CollectionExtraProperties(
+    request: Request,
+    collection: Annotated[Collection, Depends(CollectionParams)],
+) -> ExtraProperties:
+    """
+    Extracts extra properties from the first feature of a collection. (First Item has the highest priority)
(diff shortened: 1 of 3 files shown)


slesaad:
@vincentsarago can you please have a look at this? we're aware this isn't the best approach, but would love to hear thoughts on what that would be

bitner:
The entire goal of TiPg is to require no configuration to expose a feature service from Postgres/PostGIS tables without needing to add any additional functions/tables etc to the Postgres instance. If you require discovery of collections via metadata in the collections, I would recommend (and this is what we do at DevSeed) storing the metadata for each TiPG Collection as an Item in a Stac Catalog (ideally using PgSTAC).

slesaad:
Hmm understood. We thought about that too, but having a collection in two different places felt redundant. But if that's what is recommended, we'll follow that.

vincentsarago:
can we close this for now?

PR #227, opened 2025-07-01
fix rel for tiling-scheme and openapi media-type

### LINKED ISSUE #226: Returned `rel` link for tiling scheme is wrong
While I was trying to check the OpenLayers demo [https://openlayers.org/en/latest/examples/ogc-vector-tiles.html](https://openlayers.org/en/latest/examples/ogc-vector-tiles.html) out with an online TiPG deployment https://jcaujxxvec.execute-api.eu-west-1.amazonaws.com/api/docs I realize the following:

- Using `https://jcaujxxvec.execute-api.eu-west-1.amazonaws.com/api/collections/adam.adam_ts_tracks/tiles/WebMercatorQuad` as the `url` value for the `OGCVectorTile` the demo doesn't work as expected
- After a bit of debugging on the client code I have figured out that this exception is raised
   ```javascript
   throw new Error(
      "Expected http://www.opengis.net/def/rel/ogc/1.0/tiling-scheme link or tileMatrixSet"
    );
    ````
- After a further analysis on the payloads returned by TiPG I have found that the endpoints with the tiling scheme return a `rel` link with the plural form of the IRI tiling scheme `http://www.opengis.net/def/rel/ogc/1.0/tiling-schemes` while OL expects clearly the singular form `http://www.opengis.net/def/rel/ogc/1.0/tiling-scheme` according to [https://docs.ogc.org/is/20-057/20-057.html#toc30](https://docs.ogc.org/is/20-057/20-057.html#toc30) section `8.2.2` Requirement 8 Identifier D and confirmed by the exception

Image

vincentsarago:
closes #226

### DIFF
--- a/tests/routes/test_endpoints.py
+++ b/tests/routes/test_endpoints.py
@@ -63,7 +63,10 @@ def test_docs(app):
     """Test /api endpoint."""
     response = app.get("/api")
     assert response.status_code == 200
-    assert response.headers["content-type"] == "application/json"
+    assert (
+        response.headers["content-type"]
+        == "application/vnd.oai.openapi+json;version=3.0"
+    )
     body = response.json()
     assert body["openapi"]
 
--- a/tipg/factory.py
+++ b/tipg/factory.py
@@ -1254,7 +1254,7 @@ class OGCTilesFactory(EndpointsFactory):
                     "tilematrixsets",
                 ),
                 type=MediaType.json,
-                rel="data",
+                rel="http://www.opengis.net/def/rel/ogc/1.0/tiling-schemes",
             ),
             model.Link(
                 title="TileMatrixSet (Template URL)",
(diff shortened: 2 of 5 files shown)


examples/ol/main.js
```diff
@@ -0,0 +1,28 @@
...
+import VectorTileLayer from 'ol/layer/VectorTile.js';
+import OGCVectorTile from 'ol/source/OGCVectorTile.js';
+
+const map = new Map({
+  target: 'map',
+  layers: [
+    new VectorTileLayer({
+      source: new OGCVectorTile({
+        url: 'http://127.0.0.1:8081/collections/public.countries/tiles/WebMercatorQuad',
+        context: {"z": "tileMatrix", "y": "tileRow", "x": "tileCol"},
```
vincentsarago, line 13: @francbartoli do you know how it works to tell ol to use `Z/X/Y` instead of `tileMatrix/tileRow/tileCol` ?

vincentsarago:
ok I'm going to remove the example and assume the issue is fixed

PR #236, opened 2025-09-18
switch to cql2

vincentsarago:
@bitner I could need some help to finish this PR


tipg/collections.py
```diff
@@ -44,6 +43,12 @@
 TransformerFromCRS = lru_cache(Transformer.from_crs)
 
 
+def bbox_to_wkt(bbox: List[float], srid: int = 4326) -> str:
+    """Return WKT representation of a BBOX."""
+    poly = Polygon.from_bounds(*bbox)  # type:ignore
+    return f"SRID={srid};{poly.wkt}"
```
vincentsarago, line 49: This can go, we don't need to set the CRS anymore

tipg/collections.py
```diff
@@ -323,6 +328,139 @@ def queryables(self) -> Dict:
...
+        #                     bottom,
+        #                     right,
+        #                     top,
+        #                     geometry_column.srid,
+        #                 ),
+        #                 right - left,
+        #             ),
+        #             logic.V(geometry_column.name),
+        #         )
+        #     )
```
vincentsarago, line 458: TODO: Spatial filter

vincentsarago:
Note: for this PR to go through we will need to get rid of buildpg 

cc @bitner

vincentsarago:
> I'm just wondering if the switch has in its scope the validation of the CQL-TEXT/CQL-JSON in the request. If not, is this feature planned or of interest?

Yeah it should be done when we switch to CQL2 library

bkanuka:
I'm thinking of picking away at this. @vincentsarago  are you still interested in help? Anything else I might need to know?  Mostly just checking that this is still the preferred route and no one else has started work on this in the background.

vincentsarago:
@bkanuka 

> Note: for this PR to go through we will need to get rid of buildpg

This is what needs to be done 😓 

cc @bitner

vincentsarago:
done in #260

Issue #35, opened 2023-03-15
Implement OGC Features API Part 2 CRS


bitner:
Implement OGC Features API Part 2 CRS as per the specifications in https://ogcapi.ogc.org/features/

bitner:
Revisit the work in #34 to allow for using the filter-crs parameter.

PR #260, opened 2026-05-27
switch to cql2 (continuation of #236)

bkanuka:
Continuation of #236 

~~Still in draft...~~


bkanuka:
@vincentsarago  can you explain a little more about the approach to CRS ?  I don't know if the following is true:

- Data can be in any CRS, this is specified in PostGIS and we we just `ST_Transform` to 4326 at query time
- OGC Features/GeoJSON responses are _always_ in 4326
- `bbox` in the query params are always assumed to be 4326?  (or are they assumed to be the source CRS?)
- Tiles can be requested and returned in any CRS but if not specified, then assume Web Mercator EPSG:3857

Could you clarify a bit? Thanks

bkanuka:
Removed `buildpg` and building the `WHERE` with `cql2-rs`.  Ready for someone else to get eyes on this!

This is kinda a big PR (sorry).  Tests all pass.

vincentsarago:
Hi @bkanuka 
I'm trying to pick this up 

> Data can be in any CRS, this is specified in PostGIS and we we just ST_Transform to 4326 at query time

Yes, https://github.com/developmentseed/tipg/blob/1d269b344893a89177f808c7fbaa3ab4054bb5b2/tipg/collections.py#L501-L503

> OGC Features/GeoJSON responses are always in 4326

Yes

> bbox in the query params are always assumed to be 4326? (or are they assumed to be the source CRS?)

Yes, because we choose not to implement part 2 of the spec https://github.com/developmentseed/tipg/issues/35

> Tiles can be requested and returned in any CRS but if not specified, then assume Web Mercator EPSG:3857

a TMS object must be passed to the `get_tile` method https://github.com/developmentseed/tipg/blob/1d269b344893a89177f808c7fbaa3ab4054bb5b2/tipg/collections.py#L856-L860 so there no default.

bkanuka:
Great. That was my understanding, but reasoning about transformations is hard so there's a chance I fumbled it.

Side note: I'm not a fan of 1000+ line files but didn't want this pr to also include refactoring. Maybe that will come later 😉

tipg/dependencies.py
```diff
@@ -289,16 +287,29 @@ def filter_query(
...
+
+    To make that assumption explicit in the generated SQL, we round-trip
+    user input through cql2-json before re-parsing. The cql2 library emits
+    ``ST_GeomFromText(...)`` (SRID 0) when an ``Expr`` was parsed from
+    cql2-text, but ``ST_GeomFromGeoJSON(...)`` (SRID 4326) when parsed
+    from cql2-json — and SRID 0 literals fail any mixed-SRID PostGIS
+    comparison. Re-parsing the JSON form forces every spatial literal into
+    the SRID-4326 emit path. ``_where`` later wraps these 4326 literals in
+    ``ST_Transform(..., <column_srid>)`` if the target column is in a
+    different CRS, so the index on the column side is preserved.
```
vincentsarago, line 307: 👍

tipg/collections.py
```diff
@@ -323,6 +425,178 @@ def queryables(self) -> Dict:
...
+            )
+            tms_epsg = tms.crs.to_epsg() or 4326
+            tile_target_srid = target_srid if target_srid is not None else 4326
+            if tms_epsg != tile_target_srid:
+                transformer = TransformerFromCRS(
+                    tms_epsg, tile_target_srid, always_xy=True
+                )
+                west, south, east, north = transformer.transform_bounds(
+                    west, south, east, north
+                )
```
vincentsarago, line 584: I'm not sure to see where we do the transformation from TMS's CRS to the Geometry column's CRS

vincentsarago: https://github.com/developmentseed/tipg/pull/236/changes#diff-db0c25ec9d8f23d512f69d229dda0d95ed543404e78e377430e5deebeefd8c41R432-R439

tipg/collections.py
```diff
@@ -323,6 +425,178 @@ def queryables(self) -> Dict:
...
+                    east,
+                    north,
+                    srid=tile_target_srid,
+                )
+            )
+
+        if exprs:
+            # NOTE: do not call .reduce() — cql2-rs constant-folds some
+            # predicates (e.g. `"numeric" IS NULL`) incorrectly.
+            # TODO: Open a bug in cql2-rs for this.
```
vincentsarago, line 600: cc @bitner @gadomski

vincentsarago (commented):
@bkanuka This is excellent, 

Thanks for all the docstrings, it really helped for the review 🙏 

I just have one comment about the tile bbox reprojection, once we figure this we can merge 🙏

tipg/collections.py
gadomski: I can confirm:

```sh
$ cql2 '"numeric" IS NULL' --reduce
true
```

https://github.com/developmentseed/cql2-rs/issues/231

tipg/collections.py
vincentsarago: @bkanuka could you explain? 🙏

bkanuka:
I will get back to everything here next week. Still top of mind to get this merged but human life is limiting my response time ATM.

tipg/collections.py
bkanuka: `tile_target_srid` is the column CRS. So those lines _are_ where the transformation is done. I added code comments that might clarify.  (maybe I'm the one confused though!)

vincentsarago (approved):

vincentsarago:
thanks a lot @bkanuka 🙏

